PR #36, opened 2015-03-20
Retaining matrix structure

nutterb:
When you drop the exponentiated column from a one-row matrix (ie, a Cox model with only 1 continuous or binary independent variable), the object returned is a numeric vector, and `fix_data_frame` can't figure out what to do with it.  Adding the `drop=FALSE` argument keeps the matrix in tact.


nutterb:
Wow.  That was fast! :)
By the way, thank you!  I've been trying to cobble something like this together off and on for years and never quite figured out how to do it.  You've made it look so easy here.  I'll be using this a lot.

dgrtwo:
Thanks, and thanks for the fix! I'm always very interested in feature/pull requests and bug reports so please don't hesitate to inform me of any issues.

PR #42, opened 2015-04-15
Added package nlme tidiers (lme, nlme)

dmenne:
Functions for package `nlme` have been made as similar as possible to those for `lme`4`; tests and examples from lme4 are used. deviance and residuals.df are not included because I do not fully understand the definition lme4 uses.


dmenne:
Some more tests and a meaningful error message for the inconsistent case in nlme has been added.

dgrtwo:
Thanks, this is excellent work!

I've also added you as a contributor to DESCRIPTION. Let me know if there's an email address I should add to that.

PR #44, opened 2015-04-15
Loess

joranE:
Added simple augment.loess method. I didn't write any tests, but I can work on that if you'd like.


dgrtwo:
Thanks!! Please note that I made a few changes:

- Added @param descriptions of each argument
- Added newdata as an argument, passed through to augment_cols
- Added examples

PR #49, opened 2015-05-15
correct transposed coefficient matrices

nutterb:
Also returns strings instead of factors.

I had used `byrow=TRUE` when it should have been `byrow=FALSE` to construct the coefficient matrices.


nutterb:
Just wondering if the quiet on this pull request is a reflection of you being busy or a reflection of low quality coding.  Hope all is well.

dgrtwo:
Certainly not the latter! Just started a new job and devote work to broom in "chunks." Thanks for your contibution!!

nutterb:
Makes total sense. Congratulations on the job!

PR #56, opened 2015-06-11
Add a `tidy` method for x,y,z lists

jiho:
Those lists are typically used by `image` and `persp`


jiho:
Thanks!
(great work by the way; plyr + broom are excellent)

PR #60, opened 2015-07-08
boot tidiers; lme4 tidier extensions; geeglm cleanup

bbolker:
Feel free to turn down this pull request.  Unfortunately there are a few too many different things in it (I should have broken it out into three pull requests), but it's not so complicated that you can't use the pieces separately if you like.  The most interesting/complicated part is the extensions to the lme4 tidiers (which could by extension be used on a variety of mixed model tidiers), following up the discussion at https://github.com/dgrtwo/broom/issues/38 .   I'd welcome comments and opinions, particularly about the interface: are the names of the arguments appropriate?  Is this the best way to structure things?


dgrtwo:
Wow, this is great!!

It's going to take a bit of time for me to go over (sorry I haven't taken a look at your fork until now!). Most of the arguments and default values so far look good, though I'll explore lme4's confidence intervals and the FIXME's. Few quick notes:

1. DESCRIPTION contains "geeglm", shouldn't this be "geepack"?
2. Where does the "clotting" data in the boot example come from?
3. I'd probably want to return bias and std.error columns from tidy.boot, is there a reason you didn't? (I see that you noted that returning the standard error might be misleading, but I think it would be less so if the bias were returned as well).

bbolker:
1. yup. I was seeing geeglm-related warnings and tried to fix them too hastily.
2. oops again.  It's defined on the fly in `example("glm")`
```
# A Gamma example, from McCullagh & Nelder (1989, pp. 300-2)
     clotting <- data.frame(
         u = c(5,10,15,20,30,40,60,80,100),
         lot1 = c(118,58,42,35,27,25,21,19,18),
         lot2 = c(69,35,26,21,18,16,13,12,12))
```
Really any example would do here.
3. No, not particularly.  Just keeping it simple I guess.

bbolker:
Bump.  Any thoughts?  (Sorry to nag, but I'm kind of excited about this and would like to continue/write tidiers for other mixed model classes ... but no point doing that until we've hashed out more about what the interface should look like.)

dgrtwo:
Sorry! I'm working on it now; I hope to have it in the new CRAN update this weekend or Monday.

bbolker:
tiny bump.  I know you're working on this, sorry to nag.  Same point as above, really; looking forward to feedback/comments on the interface so that I can start writing tidiers for other mixed-model packages/incorporating this into my normal workflow ...

dgrtwo:
Sorry again for the delay. I actually appreciate the bumps. Today it gets done.

I'm running into a bit of confusion about the boot tidiers, particularly p-values. I'm adding comments to the code in the relevant places.

dgrtwo:
I've fixed the typo so that's not a problem; am in the process of tinkering and merging.

Right now I'm leaning towards leaving the p-values out, only because they're not mentioned in the docs or functions for the boot package itself. I like to leave the statistics to the original packages, and just rearrange values/structures/names, so that if there any bad statistics is happening it's not my fault :). For example, I'm including calculations of bias and std.error since those are given in the print.boot output. I'm also happy to call package functions like boot.ci, and if there were boot.pvalue I wouldn't hesitate.

Let's take your suggested second and third option- I'll add a `p.values` option, default FALSE, and make the issues very clear in the documentation for that option.

I'm writing this up now, along with bringing in the code for bias/stderror from print.boot (which... :-/).

dgrtwo:
After consideration I ended up leaving the p-values out; I realized that getting them right is involved enough that anyone who knows they want to do it can surely compute it themselves. And in some cases, you can compute a p-value but don't want to compare the statistics to 0 (for example, I think if you're bootstrapping residuals in a regression and adding them to the null, you'd want to compare the calculated statistics to the original statistics). So I'm leaving them out for now!

If you share an email address you'd like to use I'd be happy to add you as a "ctb" on the package.

bbolker:
Sounds good; discretion is the better part of valor.  `bolker@mcmaster.ca`.  (Let me know what you think about the mixed model stuff, which to me is the most interesting part ...)

PR #61, opened 2015-07-09
quantreg tidiers

joranE:
I took a stab at tidiers for rq, rqs and nlrq objects from the quantreg package. I would rate myself as only having an intermediate understanding of quantile regression models, so it could probably benefit from some review by an expert.

A significant difference from the "standard" output format to be aware of is that since some of these models output summary information for each quantile, I sometimes include a .tau column so you know which quantile the estimates belong to.


dgrtwo:
This is a terrific contribution and I apologize for not reviewing it until now!

I'll look over the specifics of the returned output, and maybe add examples, before submitting next version to CRAN

joranE:
No problem! As I said, I'd make sure you're happy with the output (statistically speaking) as well as the code itself.

Issue #62, opened 2015-07-23
tidy method for anova misses a column name


agoldst:
Applying `tidy` to the results of `anova` on more than one model yields a data frame where all the column names are off by one (and the last column has name `NA`):

```r
m1 <- lm(mpg ~ wt + qsec + disp, mtcars)
m2 <- lm(mpg ~ wt, mtcars)
a <- anova(m1, m2)
```

Now `print(a)` gives

````
Analysis of Variance Table

Model 1: mpg ~ wt + qsec + disp
Model 2: mpg ~ wt
  Res.Df    RSS Df Sum of Sq      F   Pr(>F)   
1     28 195.46                                
2     30 278.32 -2   -82.859 5.9348 0.007099 **
---
Signif. codes:  0 ‘***’ 0.001 ‘**’ 0.01 ‘*’ 0.05 ‘.’ 0.1 ‘ ’ 1
````

but `tidy(a)` gives

````
  df    sumsq meansq statistic  p.value          NA
1 28 195.4626     NA        NA       NA          NA
2 30 278.3219     -2 -82.85933 5.934796 0.007099496
````

It looks like `tidy.anova` assumes `a` has five columns, but it has six; the method misses the third column of `a` (the `Df` column). Apologies for not putting together a pull request, and thanks for this fantastic package.

**Comment by nutterb** [comment]
This is another type of ANOVA, and is really more of a likelihood ratio test than an analysis of variance.  

But this can be resolved as easily as adding another `else if{}`

```r
else if (identical(colnames(x), c("Res.Df", "RSS", "Df", "Sum of Sq", "F", "Pr(>F)"))){
    # x is anova(m1, m2)
    nn <- c("res.df", "rss", "df", "sumsq", "statistic", "p.value")
  }
```

PR #64, opened 2015-08-02
anova tidying changes

nutterb:
Added a logical step to identify anova objects comparing two models
(`anova(fit1, fit2)`).  Addresses Issue #62

Also added as many namespaces for `stats` as I could find.

I have not changed version numbers.

This doesn't pass `check` cleanly, but I don't think that's my fault.  Here's a list of complaints from `check`

```
tidy.glmnet: no visible global function definition for 'is'
tidy.glmnet: no visible binding for global variable 'step'
tidy.lm: no visible global function definition for 'is'
```

for the `is` functions, I tried adding `methods::is`, but got an odd note about `is` and `::` vs `:::`.  I don't have to time to fully investigate it right now, but at least a lot of the `stats::coef` etc references are properly noted now.

for the `step` issue, it looks less like the `step` function was used and more like a non-standard evaluation issue. 

```
* checking Rd \usage sections ... WARNING
Undocumented arguments in documentation object 'geeglm_tidiers'
  '...'
```

```
* checking for unstated dependencies in examples ... WARNING
'library' or 'require' call not declared from: 'geepack'
 WARNING
```

The last example in `anova_tidiers` is also generating an error.  It produced the error when I removed the code I added.  It seems with that particular model, the `colnames` is returning `NULL`.

Although not passing all of `check`, this build installs and operates.


dgrtwo:
Thanks Ben! Note that in cbee67691c998b9a1b5dd90228a85698e55e31fc I've combined your approach and mine into one using plyr::rename. I think checking for particular ANOVA types is going to keep needing special cases, but we can set up a mapping of untidy names to tidy ones.

(Exception is if there are both F statistics and Chisq statistics in the same ANOVA object. I haven't seen that yet but should watch out for it)

nutterb:
You would merge that in literally 5 minutes before I push up the rest of the changes to fix the CRAN checks.  Hold on a few minutes and I'll incorporate the `rename` changes into what I've got and that should allow for a clean merge. (yeah, I know...rebase...but that has never, ever worked for me.)

dgrtwo:
Oh dear. Yeah tonight I handle all broom business. https://github.com/dgrtwo/broom/pull/60 you're next.

PR #68, opened 2015-08-07
clean up the CRAN checks

nutterb:
How good are you at merging?  If you really need me to rebase this, I can.  Just might take me a couple days.

The highlights
Fixed up all the warnings, errors, and notes from `check`

You'll want to look at tidy.aovlist.  The example was failing, even before I started playing with it.  I think I got the output you intended based on the code that was there, but can't be too sure.

I did some hacking around in the glmnet tidiers to get the standard evaluation forms of `gather_` to work.  There's probably a cleaner way, but I couldn't think of it.

Like I said, if this isn't going to merge well, close it and I'll work on rebasing.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -47,7 +47,8 @@ Suggests:
     plm,
     biglm,
     ggplot2,
-    nnet
+    nnet,
+    geepack
 URL: http://github.com/dgrtwo/broom
 BugReports: http://github.com/dgrtwo/broom/issues
 VignetteBuilder: knitr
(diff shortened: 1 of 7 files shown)


dgrtwo:
Happy to merge, and it hasn't been difficult!

Why switch to standard evaluation? If it's because of a warning about "step", we can simply add `"step"` to `globalVariables` in globals.R.

nutterb:
The `globalVariables` thing is another piece I've never been able to make work for me (I'm typically a better analyst than I am a programmer). I won't be hurt if you go that route. In fact, it'd be nice to see how it's supposed to be done.

dgrtwo:
Added "step" [here](https://github.com/dgrtwo/broom/blob/master/R/globals.R). I get no warnings currently, though I haven't run as-cran yet. Thanks!

Issue #66, opened 2015-08-03
Update Readme.md to reflect all available tidiers


helix123:
Readme.md (frontpage of repo) does not reflect all available tidiers. At least, the tieders for `plm` (package `plm`) and `lm.ridge` (package `MASS`) could be added. Think, this is really interesting information for people looking if their model is supported by broom.

dgrtwo:
Completed with #69

PR #69, opened 2015-08-07
Automate available tidiers list

nutterb:
I'm not sure if this README.md file is where you want this, but I put in a README.Rmd file that grabs all of the classes `broom` can `tidy`, `glance`, and `augment` and sticks them in a table.  It's  a big table though, and maybe it would be better to stick this in a folder ignored by the build process.  Then you could link to it from the main `README`.

This addresses Issue #66

### DIFF
--- a/README.Rmd
+++ b/README.Rmd
@@ -1,0 +1,265 @@
+---
+output: md_document
+---
+
+broom: let's tidy up a bit
+=====================
+
+The broom package takes the messy output of built-in functions in R, such as `lm`, `nls`, or `t.test`, and turns them into tidy data frames.
+
+The concept of "tidy data", [as introduced by Hadley Wickham](http://www.jstatsoft.org/v59/i10), offers a powerful framework for data manipulation and analysis. That paper makes a convincing statement of the problem this package tries to solve (emphasis mine):
+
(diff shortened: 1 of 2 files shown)


dgrtwo:
Great idea!

I ended up moving the whole Available Tidiers section down the page a bit, since from a marketing perspective it might distract people who still don't know what the package does :).

briatte:
I'm not sure how the list is generated, but it has a row for `NULL` models, and does not list `glm`, for instance.

Question: how do you want to handle aliased methods?

Example: the `mgcv` package can estimate both `gam` and `bam` objects. As far as I can tell, the latter are just the former, but with large datasets. The results of both models are similar and should be amenable to the same `tidy` and `glance` methods.

Is it sufficient to add `tidy.bam <- tidy.gam` to the code and to alias the `gam_tidiers` documentation page with `bam_tidiers`?

nutterb:
The list is generated by making a data frame from `methods("tidy")`, `methods("glance")` and `methods("augment")`.  `glm` is missing because I used `left_join` when I should have used `full_join`.  I'll fix this when I get home tonight. Thanks for pointing my attention to that.

Believe it or not, there is a `tidy.NULL` method, which is why it shows up in the list. (https://github.com/dgrtwo/broom/blob/master/R/tidy.R).  

For `tidy.bam`, it will show up in the list as long as there is a method for `bam` objects, even if it is as simple as 

```r
tidy.bam <- function(...){
  tidy.gam(...)
}
```

briatte:
Thanks for the explanation about the `README`.

Aliasing is important for the project. Another example: the `AER` package has an `ivreg` function that can be tidied exactly like `lm`.

PR #70, opened 2015-08-13
ergm_tidiers

briatte:
Adds `tidy` and `glance` methods for `ergm` objects produced by the `ergm` package. There is no obvious way to `augment` the output of an `ergm` fit, so the method is left.

- The `tidy` function behaves like the same method for `lm`.
- The `glance` function behaves more or less like the same method for `glm`, and also covers basic MCMC settings (as an option).

I have tried to maintain consistency with names used elsewhere in the package, e.g. `AIC` and `BIC` are in uppercase, `df.residual` refers to the degrees of freedom for residual deviance, etc. I used the documentation from `lm_tidiers.R` to create my own.

Suggestion: my code computes _z_-scores in order to compute the confidence interval at any confidence level; this should probably be a utility available for other package functions.

Code passed `R CMD CHECK` without errors, notes or warnings.

### DIFF
--- a/R/ergm_tidiers.R
+++ b/R/ergm_tidiers.R
@@ -1,0 +1,198 @@
+#' Tidying methods for an exponential random graph model
+#'
+#' These methods tidy the coefficients of an exponential random graph model
+#' estimated with the \pkg{ergm} package into a summary, and construct
+#' a one-row glance of the model's statistics. The methods should work with
+#' any model that conforms to the \pkg{ergm} class, such as those
+#' produced from weighted networks by the \pkg{ergm.count} package.
+#'
+#' @return All tidying methods return a \code{data.frame} without rownames.
+#' The structure depends on the method chosen.
+#'
+#' @references Hunter DR, Handcock MS, Butts CT, Goodreau SM, Morris M (2008b).
+#' \pkg{ergm}: A Package to Fit, Simulate and Diagnose Exponential-Family 
+#' Models for Networks. \emph{Journal of Statistical Software}, 24(3). 
+#' \url{http://www.jstatsoft.org/v24/i03/}.
+#' 
+#' @seealso \code{\link[ergm]{ergm}}, 
+#' \code{\link[ergm]{control.ergm}}, 
+#' \code{\link[ergm]{summary.ergm}}
+#'
+#' @name ergm_tidiers
+#'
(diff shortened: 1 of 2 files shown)


R/ergm_tidiers.R
```diff
@@ -0,0 +1,198 @@
...
+#' @param conf.int whether to include a confidence interval
+#' @param conf.level confidence level of the interval, used only if
+#' \code{conf.int=TRUE}
+#' @param exponentiate whether to exponentiate the coefficient estimates
+#' and confidence intervals (typical for logistic regression)
+process_ergm <- function(ret, x, conf.int = FALSE, conf.level = .95,
+                         exponentiate = FALSE) {
+    if (exponentiate) {
+        # save transformation function for use on confidence interval
+        if (is.null(x$glm) ||
```
briatte: This check is probably useless here.

R/ergm_tidiers.R
```diff
@@ -0,0 +1,198 @@
...
+    s <- ergm::summary.ergm(x, ...)
+    # dyadic (in)dependence and number of MCMLE iterations
+    ret <- data.frame(independence = s$independence, iterations = x$iterations)
+    # log-likelihood
+    ret$logLik <- tryCatch(as.numeric(ergm::logLik.ergm(x)), error = function(e) NULL)
+    # null and residual deviance
+    s <- s$devtable
+    if (deviance & length(s) > 0) {
+        # dirty, but recomputing the logLik table would require much more effort
+        # that would add a dependency to the network package (for the dyad count)
```
briatte: The 'real' way to compute the following statistics requires [counting the number of dyads](https://github.com/cran/ergm/blob/master/R/summary.ergm.R#L122) in the network used by the model. This would introduce a dependency to the `network` package, so I went for the ugly hack of the deviance table instead.

dgrtwo:
This is great work!

I wouldn't mind a dependency on the network package as long as it is `SUGGESTS`, and then use `network::` to call functions in it. Since `ergm` depends on the network package, anyone using `ergm`, or installing broom with suggests included, will already have it installed.

Do you want to change your approach with that in mind? It's up to you.

briatte:
Done: `network` and `statnet.common` are now in `Suggests` (both are imported by `ergm`), and the deviance stats are computed as in `ergm::summary.ergm`.

PR #72, opened 2015-08-14
bootstrapped temporal ERGM tidier

briatte:
This really is a wrap-up of the `confint` method of the `xergm`
package. There is nothing very fancy about `term` or `mtergm` models,
and no easy-to-obtain GOF statistics, so the only method here is
`tidy`, which optionally exponentiates the coefficients and confidence
intervals.

I just noticed something: `quick=TRUE` does not take `exponentiate`
into account, even though both arguments can be passed together.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -55,7 +55,8 @@ Suggests:
     AUC,
     ergm,
     network,
-    statnet.common
+    statnet.common,
+    xergm
 URL: http://github.com/dgrtwo/broom
 BugReports: http://github.com/dgrtwo/broom/issues
 VignetteBuilder: knitr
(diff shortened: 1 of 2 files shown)


dgrtwo:
Thanks! Fixed the exponentiate issue before merging.

briatte:
I believe the `exponentiate=TRUE quick=TRUE` problem is pervasive: it also affects the tidier for `ergm`, and therefore also affects the tidier for `lm` (and probably others). The issue is that the exponentiation is done at the end:

```r
tidy.model <- function {
  if (quick)
    return(quick_stuff)

  not_quick_stuff <- ...
  return(process(not_quick_stuff, exponentiate = exponentiate, ...)
}
```

Perhaps it would work to have all `quick=TRUE` results also processed through the `process_` functions. Another solution would be to dissociate the `tidy.model` from the "quick" version, by calling the latter `qtidy` or something like that.

I hope this makes sense.

dgrtwo:
That's a really good catch!

You're right, we probably want to have a generalized tidier utility similar to the generalized `augment_cols`. I'm going to look at all the times quick leads to this problem and fix them. I probably want to keep it in the same method, though.

PR #74, opened 2015-08-14
allow quick=TRUE and exponentiate=TRUE for ergm and lm

briatte:
Here’s an easy way to solve the issue without modifying too much code.


briatte:
Thanks :)

I have not checked if other tidiers are affected, but the fix should be really easy to apply, whether or not the tidiers use an internal `process_` function.

PR #75, opened 2015-08-17
keep "method" and "alternative" when tyding htest

helix123:
Hope you find it useful. Or do you think having strings in the returned data.frame violates the tidy format?

### DIFF
--- a/R/htest_tidiers.R
+++ b/R/htest_tidiers.R
@@ -23,6 +23,8 @@
 #'   two-sample t-test}
 #'   \item{estimate2}{Sometimes two estimates are computed, such as in a
 #'   two-sample t-test}
+#'   \item{method}{Method used to compute the statistic as a string}
+#'   \item{alternative}{Alternative hypothesis as a string}
 #'   
 #' Which columns are included depends on the hypothesis test used.
 #' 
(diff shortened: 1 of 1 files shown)


dgrtwo:
Thanks for the suggestion! But I decided against returning these outputs early on. See the list of "Conventions" for `glance` output on the bottom of the README:

> We avoid including arguments that were given to the modeling function. For example, a glm glance output does not need to contain a field for family, since that is decided by the user calling glm rather than the modeling function itself.

Once we start including these columns results in glance output (and `tidy.htest` is basically glance output since it's one-value-per-model) we end up with lots of additional details from model objects, but that are basically just passed straight from the function.

If you want to compare multiple models that used different parameters for these, the way to do it would be to save the input you gave the function alongside the output. This is often easier than it sounds if you're using `group_by` and `do`. For example:

    library(dplyr)
    library(broom)

    set.seed(2015 - 08 - 17)
    x <- rnorm(10, 1)
    y <- rnorm(10)

    data_frame(alternative = c("two.sided", "greater", "less")) %>%
      group_by(alternative) %>%
      do(tidy(t.test(x, y, alternative = .$alternative)))

This gives the output

Source: local data frame [3 x 9]
Groups: alternative

      alternative estimate estimate1  estimate2 statistic   p.value parameter  conf.low conf.high
    1     greater 1.017749 0.5568465 -0.4609029  2.563225 0.0099684  17.34555 0.3278152       Inf
    2        less 1.017749 0.5568465 -0.4609029  2.563225 0.9900316  17.34555      -Inf  1.707684
    3   two.sided 1.017749 0.5568465 -0.4609029  2.563225 0.0199368  17.34555 0.1812994  1.854199

Notice that it's already saving the alternative as a column.

helix123:
Thank you for the detailed explanation! That makes perfect sense. Also, this is a nice example and very useful for me.
There are some tests which do not accept arguments like `alternative` (but have one, of course). Thus, it would not be pumping the argument from the function call to tidy's output. However, the user should know then what the (only) alternative hypothesis, so we do not need to include it in the tidied output.

PR #76, opened 2015-08-24
fix warning in tidy.anova

setempler:
Hi,

the column `term` in `tidy.anova` should be ignored in the respective `warning` call.

Best, Sven


setempler:
For completeness, the messages from the commits:

- - -

194d24e:

Column 'term' causes a warning, e.g.:

```
Warning message:
In tidy.anova(a) :
  The following column names in ANOVA output were not
  recognized or transformed: term
```

This should be ignore since the column is well defined from `fix_data_frame`.

- - -

44c9dba:

`warning` separates without space, added one to the first character string.

PR #78, opened 2015-09-01
various lme4_tidiers fixes

bbolker:
Fixed some problems with lme4 tidiers (they weren't very well tested, so when I went to actually *use* them for something ...)

Still interested in feedback about whether these design/interface decisions are the right way to go for mixed models generally (`MCMCglmm`, `glmmADMB`, `glmmML`, `Rstan`, `INLA`, ... here we come!)

### DIFF
--- a/R/lme4_tidiers.R
+++ b/R/lme4_tidiers.R
@@ -44,28 +44,24 @@ NULL
 
 #' @rdname lme4_tidiers
 #' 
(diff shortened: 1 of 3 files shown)


dgrtwo:
Still interested in looking over those too! I'm afraid I'm not nearly as experienced with mixed models as you so it takes time for me just to understand the statistical aspects.

PR #79, opened 2015-09-29
first pass at rstan tidiers

bbolker:
I'm not sure this is really ready for prime time but thought it might be worth discussion.

It's a bit hard to include rstan examples (assumes compilers set up, etc. etc.); might it be worth saving a fitted object with the package in order to be able to demonstrate tidying without running the model from scratch?


bbolker:
any thoughts about these?

bbolker:
added a summaryDefault bug fix ...

dgrtwo:
Brilliant work, sorry about the long delay!

I changed a few small things, most importantly adding an rstan example to inst/extdata so that people can play around with the tidying method. One change is to SUGGESTS coda rather than IMPORTS it; which has usually been my approach for packages needed only by a few tidying methods. (Also I made some edits punctuation/spacing, which I know I am unnecessarily fussy about).

I'm not sure what you mean by "how to make tidying methods inherit appropriately"- perhaps you could elaborate? I like your current approach of a tidyMCMC method (though I changed the package so that method is not exported).

Am going to submit this as part of broom 0.4 this weekend. I truly appreciate all of these fixes and contributions!

PR #98, opened 2015-12-04
Function to copy tidy data to the clipboard 

ahoho:
This is a function that takes tidied inputs and copies them to the clipboard, and I've found it pretty useful. I often find myself needing to paste my regression results into Excel for formatting purposes, and this wraps `writeClipboard` from `utils` to do so. 

Plays nicely with `magrittr`, in particular the [`%T>%`](https://cran.r-project.org/web/packages/magrittr/vignettes/magrittr.html) operator, which allows side effects in a chain of pipes. This is useful when you might want a record of your data in addition to copy/pasting it. For example:

```
mtcars %>% 
 lm(qsec ~ hp + wt + cyl, data = .) %>%
 tidy %T>%
 copyData %>% 
 write.csv("./Output/my_model.csv")
```


dgrtwo:
Hi, sorry for the delayed response to this issue.

I think this is a really handy function, but doesn't fit with broom, as it could be used with any data frame (or, for that matter, any matrix or vector) and not just the results of broom functions.

I think would be an excellent fit in the [clipr](https://github.com/mdlincoln/clipr) package- if you're still interested, I'd suggest submitting it there.

Thanks sincerely for the submission!

ahoho:
Thanks for the tip!

PR #101, opened 2015-12-09
Shoddy attempt to fix #100

### LINKED ISSUE #100: tidy on TukeyHSD only returns first list element instead of the whole output
Consider a more complex use case of `TukeyHSD`:

```
model <- aov(mpg ~ as.factor(gear) * as.factor(cyl), data = mtcars)
TukeyHSD(model)
```

This returns fairly large output: A list with 3 elements with very untidy data.  
Now, calling `tidy(TukeyHSD(model))`, I would expect a large `data.frame` with all test output combined, or maybe just an error, but instead I receive the tidied output of the first list element, which looks like this:

```
tidy(TukeyHSD(model))
````
|comparison |  estimate|   conf.low| conf.high| adj.p.value|
|:----------|---------:|----------:|---------:|-----------:|
|4-3        |  8.426667|  5.1878865| 11.665447|   0.0000030|
|5-3        |  5.273333|  0.9549598|  9.591707|   0.0146926|
|5-4        | -3.153333| -7.6046109|  1.297944|   0.2012687|

Possible solution:  
Run tidy on each list element, append a column specifying which variable these comparisons belong to, and then combine everything to one `data.frame`.

jemus42:
This is probably not the optimal way to do this, but I guess it's a start.

### DIFF
--- a/R/stats_tidiers.R
+++ b/R/stats_tidiers.R
@@ -122,9 +122,14 @@ tidy.spec <- function(x, ...) {
 #' @export
 tidy.TukeyHSD <- function(x, ...) {
     nn <- c("estimate", "conf.low", "conf.high", "adj.p.value")
-    fix_data_frame(x[[1]], nn, "comparison")
+    varnames <- names(x)
+    plyr::ldply(varnames, function(varname) {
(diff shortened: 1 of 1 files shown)


jemus42:
Wonderful, thanks! :)

PR #102, opened 2015-12-15
More rstan support

jgabry:
This pull request fixes a small bug and adds adds a few useful features for stanfit objects:

* Fix bug where a dimension error was thrown when dropping `droppars` if only one parameter was selected by the user.

* Make it easier to select parameters by making use of the `pars` argument to the `as.matrix` method for stanfit objects, which allows selecting vectors/matrices/arrays of parameters by group. 

* Add `rhat` and `ess` arguments (both defaulting to `FALSE`) to toggle the inclusion of Rhat and effective sample size estimates.


dgrtwo:
Excellent fixes, thanks very much! Sorry for the delayed acceptance.

PR #107, opened 2016-02-08
Adding tidiers for Mclust from mclust package

jenzopr:
Hi David,

I created tidiers for the Mclust function from the mclust package, which implements the optimal model according to BIC for EM initialized by hierarchical clustering for parameterized Gaussian mixture models.

Since this is my first PR to broom, I took the kmeans_tidier as a template and followed along.

Two issues remain:

- [ ] Is the naming fine? Should it be xx.mclust, which a lower case 'm' instead?
- [ ] In the augment function, there was an error with the fix_data_frame method while testing, so I uncommented it.

Let me know what you think.
Best,
Jens


jenzopr:
I discovered an exception in the current code when empty classes occur. Will PR the fixed version soon.

PR #108, opened 2016-02-09
Adding tidiers for Mclust from mclust package (v2)

jenzopr:
Hi David,

I created tidiers for the Mclust function from the mclust package, which implements the optimal model according to BIC for EM initialized by hierarchical clustering for parameterized Gaussian mixture models.

Since this is my first PR to broom, I took the kmeans_tidier as a template and followed along.

Two issues remain:

- [ ] Is the naming fine? Should it be xx.mclust, which a lower case 'm' instead?
- [ ] In the augment function, there was an error with the fix_data_frame method while testing, so I uncommented it.

Let me know what you think.
Best,
Jens


dgrtwo:
This is really excellent work! Sorry it took a while for me to accept it!

I made a few changes to column names to bring some in line with broom conventions. Most importantly I removed the .probability1 etc columns from the augment output (though I left .uncertainty), since I try to avoid augment having arbitrarily many columns unless it's extremely likely to be useful in that tidy form. (It's a close judgment call in this case)

Thank you for your contribution!

jenzopr:
Great that you like it! It's been a pleasure.

PR #110, opened 2016-02-13
Add tidiers for stats::prcomp

gavinsimpson:
This PR implements tidiers for `stats::prcomp` which fits principal components analysis via a SVD.

Te `tidy()` method is fully tidy, but I doubt it is useful for plotting that way; you do need the principal components as separate variables for that to work.

The PR provides `tidy.prcomp`, `glance.prcomp`, and `augment.prcomp` methods.

### DIFF
--- a/R/prcomp_tidiers.R
+++ b/R/prcomp_tidiers.R
@@ -1,0 +1,141 @@
+##' Tidying methods for principal components analysis via \code{\link{prcomp}}
+##'
+##' These tidiers operate on the results of a principal components analysis computed using \code{prcomp}. The \code{tidy} method returns a truly tidy data frame, with a variable indicating the principal omponent and another variable containing the values of all eigenvectors.
+##'
+##' The \code{glance} method returns the standard deviation, proportion variance explained, or cumulative proportion variance explained for each principal component, depending on the value of argument \code{type}.
+##'
+##' @param x an object of class \code{"prcomp"} resulting from a call to \code{\link[stats]{prcomp}}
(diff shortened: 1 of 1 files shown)


dgrtwo:
Thanks!

I'm making a few changes now to be consistent with other broom tidiers, particularly changing the column names to lowercase. Much appreciated

gavinsimpson:
Cheers @dgrtwo

dgrtwo:
I'm afraid upon playing with the tidiers for a while I found myself making rather dramatic changes. Among them:

* I'd rather not ever return both variable and sample information in the same table, especially not by default. There are times they may want to be combined, but with an `inner_join`, not an `rbind`.
* I realized that since prcomp has many similarities to the `svd_tidiers` method, the input and output should be similar to that interface. This means having `row`, `PC`, and `value` or `column`/`PC`/`value`. (These variable names are arbitrary so I tend towards consistency with previous ones)
* I removed `glance` since I don't consider one column per PC to be a tidy form (e.g. we can't plot importance on the y-axis and PC on the x axis). Instead, that data can be extracted with `matrix = "d"` or `matrix = "pcs"`, which collects all the variables. I made an exception for your augment method (basically untouched) since it's very useful to have it in this format.

Of course if you sincerely dislike my changes I can take your name off (if I'd realized all the changes I'd make I'd have proposed them before merging). I am even willing to write the methods from scratch. Again I appreciate your contribution!

dgrtwo:
Though check out this pretty cool graph I set up for the examples:

```r
pc %>%
   tidy(matrix = "samples") %>%
   mutate(region = tolower(row)) %>%
   inner_join(map_data("state"), by = "region") %>%
   ggplot(aes(long, lat, group = group, fill = value)) +
   geom_polygon() +
   facet_wrap(~ PC) +
   theme_void() +
   ggtitle("Principal components of arrest data")
```

image

PR #112, opened 2016-02-17
Grouped bootstrap

jrnold:
This PR adds an argument, `by_groups`, to `bootstrap` which if `TRUE` and the tbl has groups, will use the groups as strata when bootstrapping. 

- added this functionality to `bootstrap`
- added tests for `bootstrap`
- updated bootstrap vignette to reference the `bootstrap` function in **broom** instead including the source of the function within the vignette.


dgrtwo:
This is excellent, thank you!

PR #120, opened 2016-04-13
Added support for mgcv:gam

### LINKED ISSUE #21: Compatibility with mgcv::gam()
Current tidy broom::tidy.gam() seems to only work for package gam, and not for package mgcv. The latter does not have summary(x)$parametric.anova, but a data frame could be constructed from the output of the summary anyway.

lselzer:
Hello, I added support to mgcv:gam objects. However the output for tidy is different from gam:gam because different stats are returned by summary methods of each package. We can discuss if this can be improved. This fixes #21.


lselzer:
PD: I would add test units to this but it a nightmare to deal with the namespaces and their warning messages.

PR #122, opened 2016-04-20
Tidiers for atomic vectors

lmullen:
Sometimes functions give me a named vector (I'm thinking of the igraph centrality measures at the moment) and I want to turn that into a data frame. This lets you do that with `tidy()` in a nicer way than using `as.data.frame()` which turns names into rownames.


dgrtwo:
Awesome, good idea!

One question- how would you feel about me changing the column `x` to ` value`?

lmullen:
Not a problem at all. I'd forgotten that was the standard column name in broom.

PR #131, opened 2016-05-14
Minor updates to rstanarm tidiers

jgabry:
Fixes a few minor doc typos and tweaks the test file for better compatibility with the forthcoming rstanarm update.


benmarwick:
@jgabry I've been using `rstanarm::stan_lm` a bit using tidyverse methods, and it seems like we could make `broom` more useful for `rstanarm` users. 

I wonder if it would be possible to update along these line:

- the `tidy` method to include 5% and 95% estimate values (since we have no p-value in the `stan_lm` output, these estimate values are useful to see if the distribution includes zero or not, and a rough equivalent to decide if the predictor is significant or not)
-  the `glance` method to include the `R2` value for a model. To help those weaning themselves from `lm()` ;)
- the `augment` method to get the `pointwise` output from `loo::loo` so we can have expected log pointwise predictive density (and standard error), estimated effective number of parameters and standard error, the LOO information criterion (and standard error) for each data point. So we can have a rough analogue for Cook's Distance to measure the influence of individual data points. 

What do you think?  

I've posted [a question to Stackoverflow about getting influence measures for individual points using rstanarm](http://stackoverflow.com/questions/39578834/linear-model-diagnostics-for-bayesian-models-using-rstan). Since posting that Q and looking into the literature a bit, I reckon that an update to the augment method in broom would be an ideal solution to this problem, and maybe improve the accessibility of `rstanarm` to others using the tidyverse.

PR #140, opened 2016-06-29
Cache package dependencies

jimhester:
This will use package dependency caching on Travis, which should speed up your builds once the cache is populated.

I am glad your talk got moved to the bigger room, but sorry I missed it as a consequence :(


dgrtwo:
Thanks!

Vice versa re: your talk

PR #151, opened 2016-07-13
Add argument to `tidy.SpatialPolygonsDataFrame` to retain variables f…

jlegewie:
`SpatialPolygonsDataFrame` combine polygon information with a metadata in a `data.frame`. In many situations, it's useful to keep some of these variables. For example, it makes plotting maps much easier. My changes use the `...` argument with the same syntax as `dplyr::select` to specify the variables the user wants to keep. Obviously, a tidy SpatialPolygonsDataFrame is not a very efficient way to store this data. That is why it makes more sense that the user explicitly selects the variables she/he wants to keep. `...` is just passed to `dplyr::select` so all the helper functions work as well (e.g. `starts_with`, `ends_with` etc).

Here is some example code comparing the current version with the new version. The difference is that you don't need the `left_join` part and the type conversion of the `id` column.

```
library("dplyr")
library("USAboundaries")
library("maptools")
library("broom")
library("sp")
library("ggplot2")

# current version
shp <- us_boundaries("1900-01-01")
borders <- shp %>% tidy(region = "id_num")
borders <- borders %>% mutate(id = as.integer(id))
borders <- borders %>%
    left_join(select(shp@data, id_num, area_sqmi), by = c("id" = "id_num"))
ggplot(borders, aes(long, lat)) +
    geom_polygon(aes(group = group, fill = area_sqmi), color = "black", size = 0.1) +
    coord_quickmap() 

# with this commit
shp <- us_boundaries("1900-01-01")
borders <- shp %>% tidy(region = "id_num", area_sqmi)
ggplot(borders, aes(long, lat)) +
    geom_polygon(aes(group = group, fill = area_sqmi), color = "black", size = 0.1) +
    coord_quickmap() 
```

### DIFF
--- a/R/sp_tidiers.R
+++ b/R/sp_tidiers.R
@@ -8,7 +8,7 @@
 #'
 #' @param x \code{SpatialPolygonsDataFrame} to convert into a dataframe.
 #' @param region name of variable used to split up regions
-#' @param ... not used by this method
+#' @param ... List of variables to keep from 'SpatialPolygonsDataFrame' as comma separated list of unquoted expressions (same syntax as \link[dplyr]{select}).
 #' 
 #' @name sp_tidiers
 #' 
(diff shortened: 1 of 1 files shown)


alexpghayes:
Closing because development on `sp` tidiers is halted in favor of `sf`.

PR #157, opened 2016-08-29
Handle the lm constant model case in glance

lepennec:
Hi,

lm does not provide the fstatistic for the "constant" model ( ~ 1). With this commit, glance produces two NA_real_ for the two missing values (statistics and p.value) instead of an error.

Erwan


dgrtwo:
Thanks!

PR #165, opened 2016-09-30
Add tidiers for stats::decompose and stats::stl

atheriel:
This PR adds an `augment` s3 method for R's two built-in seasonal decompositions. It does not add `tidy` or `glance` support, since neither of these methods produce model output, per se.

The `augment` method struck me as a good for turning the output of these decompositions into something that can be added back to the data. So, for example (using the `datasets::nottem` time series):

``` r
# Perform seasonal decomposition on the data with both decompose and stl:
d1 <- stats::decompose(nottem)
d2 <- stats::stl(nottem, s.window = "periodic", robust = TRUE)

# Compare the original series to these two decompositions.
cbind(broom::tidy(nottem), broom::augment(d1), broom::augment(d2))
```

The Roxygen documentation contains the above example (demonstrating basic usage) and a much longer one for visualising both types of decomposition using ggplot2. I don't know what the policy is on including these larger examples, but I'm happy to amend this one to fit the style of the project.

Similarly, I don't know what the policy is on adding test for new tidiers, but again I'm happy to write them.


dgrtwo:
Thanks!

PR #167, opened 2016-10-18
Update boot_tidiers.R to fix confidence interval output

jbiesanz:
The confidence interval output summary (ci.tab) was transposed and the function would fail if one was bootstrapping any number of statistics other than exactly 2 (and that output was incorrectly located in the resulting data.frame). Transposing the ci.tab matrix fixes this error.

There also was an "Error: unexpected 'else' " in the middle of the function that is corrected.


janfreyberg:
+1, this should be merged ASAP

dgrtwo:
Thanks! Sorry this took a long time

PR #172, opened 2016-11-13
Update anova_tidiers.R

KZARCA:
When creating an anova object with function `drop1(x, test="Chisq")`, column `Pr(>Chi)` is not recognized


dgrtwo:
Thanks!

Issue #162, opened 2016-09-19
Updates to rstanarm tidiers 


benmarwick:
@jgabry I've been using `rstanarm::stan_lm` a bit using tidyverse methods, and it seems like we could make `broom` more useful for `rstanarm` users. 

I wonder if it would be possible to update along these line:
- the `tidy` method to include 5% and 95% estimate values (since we have no p-value in the `stan_lm` output, these estimate values are useful to see if the distribution includes zero or not, and a rough equivalent to decide if the predictor is significant or not)
-  the `glance` method to include the `R2` value for a model. To help those weaning themselves from `lm()` ;)
- the `augment` method to get the `pointwise` output from `loo::loo` so we can have expected log pointwise predictive density (and standard error), estimated effective number of parameters and standard error, the LOO information criterion (and standard error) for each data point. So we can have a rough analogue for Cook's Distance to measure the influence of individual data points. 

What do you think?  

I've posted [a question to Stackoverflow about getting influence measures for individual points using rstanarm](http://stackoverflow.com/questions/39578834/linear-model-diagnostics-for-bayesian-models-using-rstan). Since posting that Q and looking into the literature a bit, I reckon that an update to the augment method in broom would be an ideal solution to this problem, and maybe improve the accessibility of `rstanarm` to others using the tidyverse.

jgabry:
Hey Ben, yeah there's definitely room for improvement. With the initial implementation I just wanted to get it working and provide some minimal useful functionality, but there's definitely much more we can do. It's really helpful to get feedback like this because otherwise we don't know how people are using it and what would be helpful to add. So thanks! 

A few comments/questions in response to your particular points:

> the tidy method to include 5% and 95% estimate values (since we have no p-value in the stan_lm output, these estimate values are useful to see if the distribution includes zero or not, and a rough equivalent to decide if the predictor is significant or not)

Setting `intervals=TRUE` and `prob=p` when you call `tidy` should give you the lower and upper bounds of the `p`% posterior interval. So for a 90% interval (lower=5%, upper=95%) you would do
```r
tidy(fit, intervals = TRUE, prob = 0.9) 
```
Is that what you're looking for or are you referring to something different? 

> the glance method to include the R2 value for a model. To help those weaning themselves from lm() ;)

[...]

PR #177, opened 2016-12-12
Update rstanarm tidiers

jgabry:
Changes in this PR: 

* (as requested in #162) Adds `"auxiliary"` as a valid value for the `parameters` argument to `tidy.stanreg`. When set to `"auxiliary"` parameters like `sigma`, `overdispersion`, `R2`, etc. (depending on the model) will be included in the output.
* Cutpoints for `stan_polr` models now included if `parameters="non-varying"`
* For `glance.stanreg`, if `looic=TRUE` then in addition to the LOO information criterion (`looic`), the expected log predictive density (`elpd_loo`) and effective number of parameters (`p_loo`) are now also included in the output.


jgabry:
CI failures seem to be unrelated. All rstanarm related tests are passing.

PR #202, opened 2017-02-22
tidy method for objects of class nlsList (nlme package)

RemkoDuursma:
Dear package maintainers,

I have written a simple tidy method for objects of class `nlsList` (`nlme`). I hope you will find this useful and add it to the broom package. Note that there is potential overlap with issue #2, which proposed a generic solution to tidying multiple models (as `nlsList` gives a list of fitted nls models, but also a few extra methods). 
In fact I used some code from a comment in that issue to implement `tidy.nlsList`.

Remko


alexpghayes:
Hey @RemkoDuursma apologies for taking such a long time to respond to this.

We're grateful for your PR but have to pass for two reasons. The first is that tidiers for mixed models are moving to [`broom.mixed`](https://github.com/bbolker/broom.mixed), and so this PR would fit better there (#329). The second is that we're more strictly defining the behavior of tidiers to work on individual models. We now strongly recommend an explicit `purrr::map` based workflow for lists of models (#346, #353).

Thanks again for this contribution! If closing this PR puts a cramp in your workflow let us know so that we can provide a workaround or better tools.

RemkoDuursma:
No problem and thanks for that. For the record I did implement this method in the simple `nlshelper` package.

PR #203, opened 2017-03-02
Added pam (Partitioning around medoids) tidiers

jenzopr:
Hi David,

since the last PR did not get noticed (somehow ;)) , here it's again:
For one of my other projects I needed a tidy version of the cluster::pam method.
It implements the partitioning around medoids, a more robust version of kmeans.
I took the kmeans_tidier as reference and rewrote it for pam.

Let me know what you think.
Best!
Jens


jenzopr:
Will be reopened soon.

PR #206, opened 2017-03-06
Fix glance.rqs

rentrop:
Take the following example:

```
mod_rq <- quantreg::rq(speed ~ dist, cars, tau=c(0.25, 0.75))
glance(mod_rq)
```

> Error in data.frame(tau = x[["tau"]], logLik = logLik(x), AIC = AIC(x),  : 
  arguments imply differing number of rows: 2, 0

`glance.rq(mod_rq)` works just fine for single models e.g. `tau = atomic_double` but for multiple values `glance.rq(mod_rq)` fails.

### DIFF
--- a/R/rq_tidiers.R
+++ b/R/rq_tidiers.R
@@ -97,7 +97,15 @@ glance.rq <- function(x,...){
 }
 
 #' @export
-glance.rqs <- glance.rq
+glance.rqs <- function(x,...){
+    n <- length(fitted(x))
+    s <- summary(x)
+    data.frame(tau = x[["tau"]],
+               logLik = logLik(x),
+               AIC = AIC(x),
+               BIC = AIC(x,k = log(n)),
+               df.residual = vapply(s, function(x) x[["rdf"]], integer(1)))
+}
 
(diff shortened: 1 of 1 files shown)


alexpghayes:
Hey thank you for this PR! It solves the multiple tau issue in a very straightforward manner.

I talked with Dave about this and we'd really like to be strict about `glance` returning a single row. We'd prefer to throw an error in the scenario you outlined above:

```r
library(purrr)
library(broom)
library(quantreg)

taus <- c(1, 3) / 4
fit_rq <- function(tau) rq(speed ~ dist, data = cars, tau = tau)

glance(fit_rq(taus))
#> An informative error
```

and to encourage a `map` based workflow for working with multiple models:

```r
taus %>% 
  map(fit_rq) %>% 
  map_dfr(glance)
#>    tau    logLik      AIC      BIC df.residual
#> 1 0.25 -134.8775 273.7551 277.5791          48
#> 2 0.75 -130.6207 265.2414 269.0654          48
```

This is more verbose, but allows us to keep the behavior of `glance` consistent across models.

PR #214, opened 2017-03-28
Initial h2o support

kevinykuo:
Adds support for H2O GLM models (regression and binary classification)


kevinykuo:
I think CI might be failing due to regressions in other packages...

@dgrtwo could you take a look sometime? Hoping to get some feedback before I tackle the other model types.

PR #216, opened 2017-04-20
add tidy, glance and augment methods for AER::ivreg

hughjonesd:
This is my first shot at adding methods for AER::ivreg, a simple interface to run instrumental variables regressions. I'd welcome any comments. In particular, I wonder what to do about including methods for first stage regressions, and/or specific diagnostic tests (e.g. Sargan tests) for instruments.


hughjonesd:
Travis error seems to be a problem downloading nlopt, not to do with these commits.

hughjonesd:
OK, I've tried multiple times to get this to pass Travis and appveyor. It's failing but on nothing to do with my code - gmm errors. It would be great to get this change in, I use it in my own projects. Is there something I am doing wrong? I'm lost here.

dgrtwo:
Looks good and I like the examples! One question: are you able to replace the `tidy` code with just `tidy.lm(...)` called with all the arguments? It looks like the contents are almost identical, and that would reduce duplication.

hughjonesd:
I could do that. I guess the question is, in future might people want to develop `tidy.ivreg` and `tidy.lm` on separate paths - for example, adding an option to display the first stage, or if something changes in the underlying method as it did for `gmm`?

Issue #27, opened 2015-02-13
tidy lavaan output


Ohlsen:
Hi! 
It would be cool to have tidy output for the lavaan-package (https://github.com/yrosseel/lavaan) which is used for structural equation modeling and cfa.
Thanks =)

dgrtwo:
I'd also like to have that included! But I've put some time into exploring tidying methods for lavaan objects, and the objects are pretty intense. Take a look, for instance, at `getMethod(summary, "lavaan")`: there are many possible contents depending on the kind of fit, and some of them are stored in a counterintuitive way.

It's also unclear what the tidy results would be, as each latent variable model stores multiple tables. I'm sure it's possible to construct a tidied version, but I'm worried I might not be the one to do it (as I'm not familiar with the package as a user).

If you had some specific ideas for the input and output for tidying methods (or better yet, code for some of them) I'd be very happy to help incorporate them into the package!

One thing I was able to cook up was a glance method, but that's just a start:

    glance.lavaan <- function(x, ...) {
      ret <- as.data.frame(t(unclass(fitMeasures(x))))
      # rename some to fit conventions of the package
      ret <- plyr::rename(ret, c(aic = "AIC", bic = "BIC", logl = "logLik"))
      names(ret) <- gsub("pvalue", "p.value", names(ret))
      ret
    }

tjmahr:
Just chiming in to agree that `glance` would wrap `fitmeasures`, and note that `tidy` would wrap `parameterestimates`.

puterleat:
Ive added a PR which includes glance and tidy methods. Needs some thought to work out what augment should do.

DominiqueMakowski:
Any advances on this?

alexpghayes:
You can follow along at #233. Adding new tidiers is not the highest priority at the moment, so if you're hoping for new tidiers sooner rather than later the best way is to make a PR.

alexpghayes:
Merged! Small changes likely to come to the column names in `glance` before the `broom 0.5.0` release.

PR #233, opened 2017-08-01
Added basic glance and tidy for lavaan models.

puterleat:
nb. the op column is left in the tidy data frame for ease of sorting/filtering.
This addresses issue #27


alexpghayes:
I'd love to get this merged in soon if you're still interested. If you are, can you:

- [ ]  Add tests for the methods you've implemented along the lines of those in `test-lm.R`. Ideal we're looking for fairly high coverage, with the tidiers tested for a decent variety of lavaan objects
- [ ] Make all of your methods return tibbles
- [ ] Add yourself as a contributor in DESCRIPTION

benwhalley:
Cool. I've added some tests based on the `test-lm.R` file and amended the DESCRIPTION.
I have to admit, I don't fully understand the failures in Travis, which seem to be linked to rstanarm not installing. Do let me know if anything else is needed.

alexpghayes:
Sorry for taking so long to get back to you. Thanks for adding the tests!

I've made a couple changes to align with current tidyverse style. These aren't fully documented at the moment, but we've started [here](http://broom.tidyverse.org/articles/adding-tidiers.html). Couple followup requests:

- The column names in `tidy` and `glance` output should be consistent with the rest of `broom`. Can you provide descriptions of the following so I can find the best match?
    - `aic   bic   cfi    chisq  logl  npar rmsea`
    - `std.lv std.all std.nox`
- The documentation should include examples of how use `tidy.lavaan` and `glance.lavaan`. It should also fully describe what each column in the output represents.
- For `glance` I think it would be better to remove the `fit.measures` argument and always return the largest reasonable set of measures of fit. Users can then later remove the ones they aren't interested in with `select`.

The Travis failures are currently on our end, we're currently only asking PRs to pass the AppVeyor build.

benwhalley:
Just to respond inline below...

> The column names in tidy and glance output should be consistent with the rest of broom. Can you provide descriptions of the following so I can find the best match? aic bic cfi chisq logl npar rmsea

logl is log likelihood
chi2 is the chi squared — a measure of discrepancy between data and model fit
aic, bic are measures of fit based on information theory
npar is number of params in the model
rmsea is root mean squared error of approximation — another fit measure

std.lv std.all std.nox: these are all different types of standardised coefficient. That is, the params are standardised with respect to the latent variables only (std.lv) or all, or just excluding exogenous variables (nox)

On reflection I wonder if it would be better to offer this dataframe unstandaqdized by default, and in long form, with an additional `term` variable if the user requests standardised coefs?

> The documentation should include examples of how use tidy.lavaan and glance.lavaan. It should also fully describe what each column in the output represents.

I can add this

> For glance I think it would be better to remove the fit.measures argument and always return the largest reasonable set of measures of fit. Users can then later remove the ones they aren't interested in with select.

The most common case here, I think, is to want to compare models on various common fit indices, and perhaps also check the number of params. in the model, and the number of observations. On reflection, I think it might be best if I added the model N, and the estimator to this and always returned in long form. There's actually no point replicating lavaan::fitmeasures too closely... the trouble there is that one always has to subset the fitmeasures returned by default, and look elsewhere for the N and estimator.  I'll the another look

alexpghayes:
Thanks! I'll probably change some of the `glance` column names up before the release for consistency with the rest of `broom`, but this looks great! I appreciate your contribution!

PR #234, opened 2017-08-01
Added index column to tidy_mcmc; corrected deprecated Stan syntax

dmenne:
Following Brian Ripley, it the definition of deviance is problematic for any lme model, not only for REML. 
Currently, a warning is produced even if there is a special case in utilities.R
The documentation always noted that NA is returned for deviance. This is no enforced both for REML and ML. 

Test for ML was added

### DIFF
--- a/man/mcmc_tidiers.Rd
+++ b/man/mcmc_tidiers.Rd
@@ -85,6 +85,9 @@ if (require("rstan", quietly = TRUE)) {
   ggplot(tds, aes(estimate, term)) +
     geom_errorbarh(aes(xmin = conf.low, xmax = conf.high)) +
     geom_point(aes(color = method))
+  
+  # Show indices as columns
+  tidy(rstan_example, conf.int = TRUE, pars = "theta", index = TRUE)
 }
 
 


dmenne:
Corrected documentation of index in mcmc_tidiers

dmenne:
Optionally with parameter index, add two columns in tidy_mcmc, so that in additions to a[3] there are columns a and 3.
Also corrected syntax in Stan file (<- is deprecated)

bbolker:
Are you willing to submit this pull request/make these changes to https://github.com/bbolker/broom.mixed ...  ? (If necessary I could re-do them by hand but ...)

dmenne:
Hi, Ben, Willing: yes. Able: ahem, I am not a great githubber. Do I have to clone your's, test the changes with your clone and submit patch? Or is there a shortcut?

bbolker:
Unfortunately I don't know of a short cut.  If this is too much of a pain I can get around to doing it too ...

dmenne:
Will do it. I owe you so much....

dmenne:
After checking, I think it's better I don't do this, because you have already made quite a few changes in your code, and I don' t want to step on your toes. Take anything you like!
Looks like a good ideas, though, broom core has grown into baroque shape lately.

alexpghayes:
Unfortunately true! Has this been incorporated in `broom.mixed` yet? I'd like to close this since we're deprecating the mixed models tidiers in favor of `broom.mixed` (#329), but want to make sure that doesn't result in lost work.

PR #242, opened 2017-09-29
Unbreak tidy.gmm(confint = TRUE) for newer versions of gmm

hughjonesd:
Existing code breaks `example(glance.gmm)` because `gmm` decided to create a new kind of confint object. This hacks tidy.gmm to avoid the problem. Maybe now we can unblock the pull request queue.


dgrtwo:
Thanks! Sorry, I've been extremely behind on merging, not your fault.

PR #248, opened 2017-10-11
add tidy, glance and augment for speedglm::speedlm models

hughjonesd:
`speedglm` is a quick and easy way to handle big data. The `tidy` method works exactly the same as tidy.lm, and simply reuses that function. `glance` and `augment` are slightly different. If you want, I could manually work out `.residuals` for `augment`; at the moment, it only returns `.fitted`.


dgrtwo:
Thanks!

PR #251, opened 2017-11-02
Add muhaz tidiers

adibender:
`muhaz` is a package which provides kernel-based, non-parametric hazard rate estimates. Right-censoring is allowed.


dgrtwo:
Thanks!

PR #252, opened 2017-11-03
add support for emmeans, the successor to lsmeans

mjskay:
[emmeans](https://github.com/rvlenth/emmeans) is now the "official" successor to `lsmeans` --- it is from the same author as `lsmeans` and has nearly the same interface, but with a more generic name ("estimated marginal means" instead of "least-squares means"). According to the author of `emmeans`/`lsmeans`, `lsmeans` will eventually be retired.

This pull request:

- Extends the existing code that supports `lsmeans` output (the `ref.grid` and `lsmobj` classes) to also support the `emmGrid` class from `emmeans`. 

- Changes the examples to use `emmeans` and renames files and the help page to start with `emmeans` instead of `lsmeans`, to reflect that `emmeans` is now the preferred name and package.

It does not remove support for `lsmeans`, on the assumption that plenty of old code still uses it.


dgrtwo:
Thanks!

PR #255, opened 2017-11-10
Logit

philmikejones:
I would like to include pseudo-R^2 and p.values for logit models, which I have added to the `finish_glance()` function.

I've added error catching but the new code fails when testing an ordinary `lm()` model. I can't tell if this is my new code, or errors already in the build.

If you have time could you help me incorporate these extra bits of logit information?


alexpghayes:
Hey @philmikejones sorry for taking so long to get back to you, and thanks for the PR! Are these additional measures well defined for other GLMs? Based on the code it looks like they'll also be calculated for poisson, etc, regressions as well as logistic regressions.

I'm also curious how widespread these measures are. `tidy.glm` is used fairly heavily and so adding these may be confusing if their use is discipline specific, for example.

PR #257, opened 2017-11-13
df in glance for lm and aov objects

### LINKED ISSUE #212: glance on an aov model reports the wrong df value
Unless I'm misunderstanding something, glance should report the same df as in summary (e.g. 2 in this example) but it seems to always report df+1

```{r}
library(broom)

d <- data.frame(
  y = c(3,4,3.5,4.1,6,7),
  group = c("a","b","a","b","c","c")
)
m <- aov(y~group,data = d)

summary(m)
glance(m)
```

This is with broom 0.4.2 (the current version with tidyverse if I'm not mistaken)

lemna:
Use numerator df from the `fststatistic` slot - fixes #212.


alexpghayes:
I know it's been forever, but I'm finally ready to pull the trigger and get this change into `broom`. Would you mind resolving the conflict and then I'll merge?

alexpghayes:
Ended up doing this in #624 because it was easier than figure out how to make git happy after file renaming here. Thanks for this!

PR #261, opened 2017-11-20
added support for likelihood-ratio chi-squared from package `car`

mariusbarth:
Hi!
When working with generalized linear models, I figured out that broom does not support output from likelihood-ratio tests that are generated by `car::Anova()`, yet. I added this tiny feature so that it works on my machine -- however, I'm not aware if there is more work to do to implement this. If it is, I will be happy to do it.


alexpghayes:
Hmmm. Interesting that `car::Anova` still returns an `anova` object. Would you add a quick test for the `car::Anova` case and update the documentation to reflect support for `car::Anova` output?

PR #262, opened 2017-11-20
Add tidiers for the loo package

wjakethompson:
This pull request adds tidiers for the `loo` class of objects, exported from the [loo package](https://cran.r-project.org/web/packages/loo/index.html). This idea, and much of the heavy lifting, originally came from @jrnold over at stan-dev/loo#48. I edited slightly to return `data.frame` objects instead of a `tibble` to be consistent with the rest of the `broom` package. I also added documentation and tests for the tidiers.


alexpghayes:
I'm sorry to have let this go so stale. Now that `broom.mixed` is handling Bayes, etc, this probably belongs there. Closing and will open a new issue there.

PR #269, opened 2017-12-30
test lmerMod via inherits rather than class

### LINKED ISSUE #125: lme4 parameters changed
Hello.

When I use glance with a lme4 model I get this warning:

> Warning message:
> In deviance.merMod(x) :
>   deviance() is deprecated for REML fits; use REMLcrit for the REML criterion or deviance(.,REML=FALSE) for deviance calculated at the REML fit.

bbolker:
I think this will resolve #125 ... the problem is with testing objects created by the `lmerTest` package; since `lmerTest` uses S4 classes, these fits have `class(fit)` = "merModLmerTest".   This *inherits from* `lmerMod`, but the string "lmerMod" is nowhere to be found in `class(x)` ... `inherits()` is a good test to discover whether the fit is a linear mixed model.

```
lmm1 <- lmer(Reaction ~Days+(Days|Subject),sleepstudy) 
inherits(lmm1,"lmerMod") ## TRUE
glmm1 <- glmer(round(Reaction) ~Days+(Days|Subject),sleepstudy,family=poisson) 
inherits(glmm1,"lmerMod") ## FALSE
```


alexpghayes:
Awesome, thanks.

Issue #249, opened 2017-10-12
use texreg::extract for tidy


hughjonesd:
The `texreg` package extracts coefficients and goodness-of-fit statistics from 94 models, many of which aren't yet covered by broom. It would be easy to write a connector that outputs `tidy` results from `texreg::extract` results. Doing the same for `glance` would also be possible. Would you be open to this? If so, I can submit a pull request.

dgrtwo:
This is a good idea! Now that I'm back to active maintenance of broom, would greatly appreciate a pull request.

hughjonesd:
OK! It might also be worth talking to Philip Leifeld (@leifeld) about future directions. (And for that matter @leeper whose `prediction` package covers some similar areas...)

alexpghayes:
Closing, details in #281.

PR #281, opened 2018-02-18
Try fallback to texreg

hughjonesd:
This is a first shot at using `texreg::extract` as a fallback. It changes the default tidy method. I haven't yet tried to do `glance`, want to see what you think of this approach first.

texreg itself falls back to `broom`; the code tries to check that this won't happen.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -91,7 +91,8 @@ Suggests:
     AER,
     muhaz,
     speedglm,
-    tibble
+    tibble,
+    texreg
 URL: http://github.com/tidyverse/broom
 BugReports: http://github.com/tidyverse/broom/issues
 VignetteBuilder: knitr
--- a/R/nls_tidiers.R
+++ b/R/nls_tidiers.R
@@ -56,7 +56,7 @@ tidy.nls <- function(x, conf.int = FALSE, conf.level = .95,
                      quick = FALSE, ...) {
     if (quick) {
         co <- stats::coef(x)
-        ret <- data.frame(term = names(co), estimate = unname(co))
+        ret <- data.frame(term = names(co), estimate = unname(co), stringsAsFactors = FALSE)
         return(ret)
     }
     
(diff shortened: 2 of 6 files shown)


hughjonesd:
@leifeld this is a temporary fix... best solution involves more collaboration and code sharing! But it seems a useful step for now.

alexpghayes:
Hey there. This looks interesting, but I'm worried about silent failures. In particular, what if a model object has a `coef` field but those coefficients don't mean what the user thinks they mean, or the naming is non-standard?

Do you have some example use cases that we could consider?

hughjonesd:
This is ultimately a matter of how much you trust texreg objects. That
package is designed for creating tables of regressions, and is quite
battle-tested, so the coef field is pretty certain to do what you'd expect
(in that sense it is probably more tightly defined than the output of
`tidy`, which makes guarantees of form rather than semantics, if I
understand it right).

Use cases are defined by the set of objects for which there is a texreg
method but not a tidy method yet!
-- 
Sent from Gmail Mobile

alexpghayes:
@dgrtwo you were pretty enthusiastic about this in #249. I don't really know enough about `texreg` to make this call.

alexpghayes:
Hey @hughjonesd I talked with Dave about this yesterday. We think this is a really cool idea but don't think this fits with the current direction of broom development. We'd like to make `broom` more consistent, part of that being that we'd like to explicit test all the tidiers that we do export, and to informatively fail for model objects that we don't export.

This is unfortunate in the case of this PR because `texreg` supports many more models then `broom` currently does. For those that are most important to your workflow, please open issues so that we can write tidiers for them sooner rather than later!

hughjonesd:
No, I understand and think it's a reasonable decision. It might well be a
good idea to do some manual copying-over of the texreg code. A lot of it
might work with just the obvious adjustments. @pleifeld might help but he
is very busy nowadays!
David

PR #298, opened 2018-04-05
Compatibility with new rstanarm and loo packages

jgabry:
This PR makes a small change to provide compatibility with the upcoming releases of rstanarm and loo. All tests pass for me locally.


alexpghayes:
Thanks!

Issue #296, opened 2018-03-30
Is the psych dependency necessary?


WillemSleegers:
I checked the code to see why the `psych` package is a dependency, and I only see a function being used in one of the examples and in a test file.

nutterb:
It is used in [`tidy.data.frame`](https://github.com/tidyverse/broom/blob/master/R/data.frame_tidiers.R#L57).

WillemSleegers:
Weird, it doesn't show up if you search for psych.

Seems a bit silly to require an entire package for a simple describe function, though. Each of the summary statistics can be calculated relatively easily.

dgrtwo:
[This](https://github.com/tidyverse/broom/blob/defdf42737e06b22f802b7c24df5d82d4a528ac1/R/data.frame_tidiers.R#L51) showed up for me in a search

Yes, I added that a long time ago and I'd love to remove the dependency; I'd merge a PR for that.

dgrtwo:
(FWIW I made the decision to depend on psych way back here: https://github.com/tidyverse/broom/pull/7 )

nutterb:
Not saying this is the right/best solution, but this is what I put together when I was playing with the `broom.base` idea (near-zero dependencies for objects in `base` R.

https://github.com/nutterb/broom.base/blob/master/R/data.frame_tidiers.R

raymkchow:
Removing psych::describe as the dependency is good, especially when you found the psych::describe output (the column names) is not so tidy when `quant` is a 1-length vector.

alexpghayes:
@nutterb `broom.base` and similar are high priorities for discussion later in the summer. At the moment, however, we're just doing a little bit of maintenance.

If you'd like to create a PR with just the `data.frame` tidier code from `broom.base`, I'd be happy to accept it. If you don't have time for that and don't care about having that in your git history, let me know if you're okay with me just swiping the code.

The data frame tidiers are going to be deprecated in the next release, but this will let us get rid of the `psych` dependency for that release.

PR #313, opened 2018-06-04
Remove `psych` dependency from data frame tidiers

nutterb:
* Revamps the `tidy.data.frame` method to directly calculate the metrics previously calculated by `psych`.  
* Removes `psych` package from the `Imports` in `DESCRIPTION`.
* I'm not completely clear on the protocol here, but I pretty blatantly borrowed from the `moments` package to get calculations for skew and kurtosis.  I've added that source to the references in the documentation for `tidy.data.frame` and added the authors of `moments` to the authorship list in `DESCRIPTION`.  

See Issue #296


alexpghayes:
Awesome. Thanks!

PR #142, opened 2016-07-07
rms tidiers

nutterb:
Includes methods:
- `tidy.cph`
- `tidy.lrm`
- `tidy.ols`
- `tidy.anova.rms`,
- `tidy.summary.rms`
- `tidy.validate`
- `glance.rms`

This mostly satisfies Issue #30 but notably does not include `augment.cph`, `augment.lrm`, or `augment.ols`.  I did not write these because 1) I'm lazy, and 2) I'm not sure what changes may be coming out due to #137, and would prefer to only write it once.

A `tidy.calibrate` might be considered in the future, but I haven't put much thought into what that would look like.

Oh, and I moved `zoo` from Suggests to Imports

PR #326, opened 2018-06-07
[WIP] Work on tidiers for rms package

nutterb:
This is a duplicate of #142, which may now be cancelled.


alexpghayes:
Mind if I play with this a bit and see if I can bring it back to life over the next couple days?

nutterb:
You're welcome to it.  I'm not likely to get around to anything outside of my real job for quite some time.

PR #334, opened 2018-06-08
Keep residuals for manova fits issue 307 Feature request

jarvisc1:
Removed a row that was deleting the residual information. 
Changed the test to include the extra row. 
I tried devtools::load_all() and it worked.
Then I installed from my github repo and it also worked.


alexpghayes:
Way to go!

Re: finding functions: In RStudio try `Ctrl` + `.` and then type a function name to find the file it lives in!

PR #341, opened 2018-06-09
Add tidiers for car::durbinWatsonTest objects

### LINKED ISSUE #266: glance method for object of class DurbinWatsonTest
I think I can write this myself, but it would be helpful to have a `glance` method for an object of class `DurbinWatsonTest` from the `car` package. The return value has the following structure:
```
List of 4
 $ r          : num 0.0624
 $ dw         : num 1.85
 $ p          : num 0.304
 $ alternative: chr "two.sided"
 - attr(*, "class")= chr "durbinWatsonTest"
```

mkuehn10:
Trying to resolve #266


alexpghayes:
You need to add `car` to **Suggests** in `DESCRIPTION`. Probably the best way to do this is with the `usethis` package: `usethis::use_package("car", type = "Suggests")` I believe. Then you'll want to use `library(car)` in the test file, or refer to the test function with a `car::` prefix.

Does that help?

alexpghayes:
Also I think it would be best to just implement a `glance` method, and no `tidy`. `tidy` is meant to be used on model components, but here we can only really summarize the whole model, so `glance` is the way to go.

mkuehn10:
I noticed that the htest tidiers make the tidy and glance the same, so that's what I did for this one as well.

alexpghayes:
Okay, that's fine then!

mkuehn10:
@karldw I think this resolves #266

Issue #270, opened 2018-01-12
tidy and augment Methods for caret::confusionMatrix


benwhalley:
How about something like this:

```
cm <- caret::confusionMatrix(rbinom(100,3,.5),rbinom(100,3,.5))

tidy.cm <- function(cm){
  cml <- cm$overall %>% as.list()
  data_frame(statistic = c("Accuracy", "Kappa"), 
             estimate = c(cml$Accuracy, cml$Kappa),
             lower = c(cml$AccuracyLower, NA),
             upper = c(cml$AccuracyUpper, NA),
             p.value = c(cml$AccuracyPValue, cml$McnemarPValue))
}

augment.cm <- function(cm){
  cm$byClass %>% as.data.frame()
}


tidy.cm(cm)
augment.cm(cm)
```

Do people have feelings about the col names for augment?

amcadie:
@benwhalley I was going to open this very issue, but you beat me to it!

One tweak could be to add a %>% t( ) to the augment.cm function so that it is easy to rbind output from several models.

Thanks for suggesting this!

randomgambit:
hello there and thanks for this life-saving package!! is `broom` able to work with `caret` natively? thanks!

alexpghayes:
Sounds good. You'll want to start by forking the broom repository, and then cloning it to your computer. Once you have it on your local computer, you can start create a new file (probably call it `R/caret_tidiers.R`). Then you'll want to add the functions defined above, except make sure they return tibbles.

After you've get the functions working the way you want, you'll want to [document their behavior](http://r-pkgs.had.co.nz/man.html).

Then next step will be to [write some tests](http://r-pkgs.had.co.nz/tests.html) to ensure that your functions are working correctly. You'll want to create a file `tests/testthat/test-caret.R` and then write the tests there.

Once the tests are passing, the final step is to commit all your work to your fork, and then open a pull request to `broom`.

A good general reference is Hadley Wickham's book R Packages, which I've linked to above.  You should also take a look the [vignette on new tidiers](http://broom.tidyverse.org/articles/adding-tidiers.html).

It's easy to get information overload, so just drop me a question anytime!

mkuehn10:
Does only a tidy function (and possibly a glance function) make sense for a confusionMatrix?  

[...]

PR #344, opened 2018-06-10
add caret tidiers

mkuehn10:
implements #270


mkuehn10:
Failure message was
`Packages suggested but not available: 'gamlss' 'ks' 'rgeos'
The suggested packages are required for a complete check.
Checking can be attempted without them by setting the environment
variable _R_CHECK_FORCE_SUGGESTS_ to a false value.
See section 'The DESCRIPTION file' in the 'Writing R Extensions'
manual.`

I don't think my PR has anything to do with those packages,  I added `caret` to the DESCRIPTION for Suggests.

R/caret_tidiers.R
```diff
@@ -0,0 +1,111 @@
+#' Tidying methods for confusionMatrix objects
+#'
+#' Tidies the result of confusion matrix from the \code{caret} package.
+#' Only a \code{tidy} method is provided, not an \code{augment} or
+#' \code{glance} method.
+#'
+#' @param x An object of class \code{confusionMatrix}
+#' @param show_class A boolean of whether to show the values for class specific
```
alexpghayes: I'd use "logical" instead of boolean since that's more standard for R.

alexpghayes: I'd also rename `show_class` to `by_class`.

R/caret_tidiers.R
```diff
@@ -0,0 +1,111 @@
...
+#' quantities from the confusion matrix (specificty, sensitivity, etc.). If set
+#' to FALSE, result will only show accuracy and kappa.
+#' @param ... extra arguments (not used)
+#'
+#' @return A data.frame with one or more of the following columns:
```
alexpghayes: Should return a tibble!

mkuehn10: I think I need to fix the durbinWatsonTest as well then.

mkuehn10:
It's giving me errors on my test script saying that str_to_lower couldn't be found, but it's from the stringr package.

atyre2:
WRT to builds failing on travis-ci, this message is at the end of all the failures I"ve just checked:
The job exceeded the maximum time limit for jobs, and has been terminated.
It looks like it is taking a very long time to get the machine set up, so never gets around to testing the package.

alexpghayes:
At the moment it's okay for the Travis build to fail, that's on us, but AppVeyor does need to pass. Pretty busy the next couple of days but will get back to this at some point this week.

mkuehn10:
Thanks @atyre2 I saw that as well.  Thanks for the response @alexpghayes, I'll do whatever is needed to fix this, but I'm really not sure what's going on.

alexpghayes:
Awesome! Thanks for your persistence with this!

PR #346, opened 2018-06-11
glance method for list objects

ellessenne:
Hi,
this is more of a proof of concept than a full featured pull request, I am posting mainly to ask for feedback. I was thinking that having a `glance` method for list objects could be useful, consider for instance the following scenario.
Say you have a given dataset with a non-linear relationship between exposure and outcome:

```r
# Simulate some non-linear data
set.seed(1)
n <- 30
x <- rnorm(n)
e <- rnorm(n)
y <- sinpi(x) + e
df <- data.frame(y, x)
```

Say you want to fit several models with different functional forms:

```r
# Fit several models
fit.list <- list(
  "fit.d01" = lm(y ~ x, data = df),
  "fit.d02" = lm(y ~ poly(x, 2), data = df),
  "fit.d03" = lm(y ~ poly(x, 3), data = df),
  "fit.d04" = lm(y ~ poly(x, 4), data = df),
  "fit.d05" = lm(y ~ poly(x, 5), data = df),
  "fit.d06" = lm(y ~ poly(x, 6), data = df)
)
```

Say you want to compare all models at once, and then select the one that fits the data best via information criteria such as AIC and BIC. You could code this very easily, using `lapply` or something similar, but it would also be nice to have a `glance.list` method that calls the appropriate method for each model in the list and returns a data frame with one row per model! In that case, you would get something like:

```r
# Compare fit
glance(fit.list)
   r.squared adj.r.squared     sigma statistic     p.value df    logLik      AIC      BIC deviance df.residual       x
1 0.04179179   0.007570073 0.9624715  1.221207 0.278529688  2 -40.38574 86.77148 90.97507 25.93784          28 fit.d01
2 0.14642060   0.083192501 0.9250752  2.315752 0.117975809  3 -38.65134 85.30268 90.90747 23.10563          27 fit.d02
3 0.38980757   0.319400754 0.7970465  5.536503 0.004474091  4 -33.61638 77.23275 84.23874 16.51736          26 fit.d03
4 0.40582954   0.310762271 0.8020888  4.268867 0.009063211  5 -33.21726 78.43451 86.84169 16.08366          25 fit.d04
5 0.41155195   0.288958600 0.8146769  3.357050 0.019361757  6 -33.07209 80.14418 89.95256 15.92876          24 fit.d05
6 0.52173888   0.396975105 0.7502492  4.181814 0.005501251  7 -29.96211 75.92423 87.13381 12.94610          23 fit.d06
```

This would make models comparison straightforward, with all the benefits of `broom`!
If you guys think this would be interesting, I could polish this up a bit; regardless, I am happy to discuss this further!
Cheers,

Alessandro

### DIFF
--- a/R/glance.R
+++ b/R/glance.R
@@ -14,5 +14,35 @@ glance.NULL <- function(x, ...) data.frame()
 
 #' @export
 glance.default <- function(x, ...) {
-  stop("glance doesn't know how to deal with data of class ", class(x)[1], call. = FALSE)
+  stop("glance doesn't know how to deal with data of class ", class(x), call. = FALSE)
+}
+
+#' @export
+#' @example
+#' # Simulate some non-linear data
+#' set.seed(1)
+#' n <- 30
+#' x <- rnorm(n)
+#' e <- rnorm(n)
+#' y <- sinpi(x) + e
+#' df <- data.frame(y, x)
+#' # Fit several models
+#' fit.list <- list(
+#'   "fit.d01" = lm(y ~ x, data = df),
+#'   "fit.d02" = lm(y ~ poly(x, 2), data = df),
+#'   "fit.d03" = lm(y ~ poly(x, 3), data = df),
+#'   "fit.d04" = lm(y ~ poly(x, 4), data = df),
+#'   "fit.d05" = lm(y ~ poly(x, 5), data = df),
+#'   "fit.d06" = lm(y ~ poly(x, 6), data = df)
+#' )
+#' # Compare fit
+#' glance(fit.list)
+glance.list <- function(l, ...) {
+  out <- lapply(seq_along(l), function(i) {
+    x <- glance(l[[i]])
+    x$x <- names(l)[i]
+    x
+  })
+  out <- do.call(rbind.data.frame, out)
+  out
 }


alexpghayes:
I like the idea and am glad you're thinking about things like this!

I'd prefer not to add this to broom for two reasons:

1. We'd prefer that users think of `broom` verbs as acting on single models, and we want people to use `purrr::map` based workflows to deal with multiple models at once. For example: `purrr::map_df(fit.list, broom::glance)` is a little less magical.
2. `tidy.list` works with a single model at a time, so these would be inconsistent.

Thanks for thinking about stuff like this! Are you using workflows like this that need some polishing in `broom`?

ellessenne:
I see, thanks for the feedback! It's not that big of a deal, as you say it is enough to use `purrr::map`, `lapply`, or anything like this really!

I thought about this as I have been doing lots of model comparisons (same model family, different formulation) picking the one that fitted the data best via AIC/BIC! I suppose it should be a quite common scenario? I wonder, would it be worth it to make it clear (say in a vignette) that `broom` verbs act on single model and that the `purrr::map` workflow is the recommended way of dealing with multiple models at once? Thanks again, and please go ahead and close this pull request whenever necessary!

alexpghayes:
That's a fantastic idea! Opening issue for that now!

PR #355, opened 2018-06-12
fix durbinWatsonTest tidier to return a tibble

mkuehn10:
addendum to #266


R/durbinWatsonTest_tidiers.R
```diff
@@ -3,7 +3,7 @@
 #' Tidies Durbin-Watson Test objects, from the \code{car} package
 #' into a one-row data frame.
 #'
-#' @param x An object of class \code{"durbinWatsonTest"}
+#' @param x An object of class \code{durbinWatsonTest}
 #' @param ... extra arguments (not used)
 #'
```
alexpghayes: Just switched over to documenting with rmarkdown, this seems like a good idea to fix now as well: [`rmarkdown`](https://cran.r-project.org/web/packages/roxygen2/vignettes/markdown.html).

R/durbinWatsonTest_tidiers.R
```diff
@@ -25,7 +25,8 @@ NULL
 #' @rdname durbinWatsonTest_tidiers
 #' @export
 tidy.durbinWatsonTest <- function(x, ...) {
-  fix_data_frame(
+  df <- 
+    fix_data_frame(
```
alexpghayes: It's only necessary to call `fix_data_frame` on `data.frame` objects. In this case it's better to just directly create a `tibble`.

R/durbinWatsonTest_tidiers.R
mkuehn10: I'm not sure what needs to be done for this.

alexpghayes: Instead of using \code{car}, use `car`. It's  a different way to markup the documentation.

PR #357, opened 2018-06-12
Smarter binary packages on Travis

alexpghayes:
Courtesy of Jim Hester. https://gist.github.com/alexpghayes/750b94342d26c49bc37dac12e6f8ffaa

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,19 +1,18 @@
-# Sample .travis.yml for R projects
-
 language: r
-sudo: false
+sudo: required
 cache: packages
 
 r_binary_packages:
-  - lme4
+  - rstanarm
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -1,19 +1,18 @@
...
+  - rstan
   - dplyr
-  - purrr
-  - devtools
-  - git2r
-  - data.table
-  - stringi
-  - matrixStats
-  - StanHeaders
+  - Matrix
```
jimhester: Matrix is a recommended package, so there is no binary, you will need to remove this

jimhester: I would try with *just* `-rstan` and `-rstanarm`

Issue #299, opened 2018-04-06
broom::tidy isn't working with within-subject ANOVA outputs


IndrajeetPatil:
Hi, I am working with the latest version of broom (`0.4.4`) and having an issue while getting the tidy output. Here is a fully reproducible example:

``` r
# loading the necessary libraries
library(tidyverse)

# converting the iris dataset to long format
iris_long <- datasets::iris %>%
  dplyr::mutate(.data = ., id = dplyr::row_number(x = Species)) %>%
  tidyr::gather(
    data = .,
    key = "condition",
    value = "value",
    Sepal.Length:Petal.Width,
    convert = TRUE,
    factor_key = TRUE
  ) %>%
  tidyr::separate(
    col = "condition",
    into = c("attribute", "measure"),
    sep = "\\.",
    convert = TRUE
  ) %>%
  tibble::as_data_frame(x = .)

# specifying the model
iris.mod <-
  stats::aov(formula = value ~ attribute * measure + Error(id / (attribute * measure)),
             data = iris_long)

# checking the summary
summary(iris.mod)
#> 
#> Error: id
#>           Df Sum Sq Mean Sq F value Pr(>F)
#> Residuals  1    264     264               
#> 
#> Error: id:attribute
#>           Df Sum Sq Mean Sq
#> attribute  1  237.2   237.2
#> 
#> Error: id:measure
#>         Df Sum Sq Mean Sq
#> measure  1   1114    1114
#> 
#> Error: id:attribute:measure
#>                   Df Sum Sq Mean Sq
#> attribute:measure  1 0.7993  0.7993
#> 
#> Error: Within
#>                    Df Sum Sq Mean Sq  F value Pr(>F)    
#> attribute           1  470.1   470.1 1445.988 <2e-16 ***
#> measure             1   57.7    57.7  177.458 <2e-16 ***
#> attribute:measure   1    1.5     1.5    4.725 0.0301 *  
#> Residuals         592  192.5     0.3                    
#> ---
#> Signif. codes:  0 '***' 0.001 '**' 0.01 '*' 0.05 '.' 0.1 ' ' 1

# tidying the ouput
broom::tidy(iris.mod)
#> Error in if (attr(object$terms, "intercept")) table <- table[-1, ]: argument is of length zero
```

Created on 2018-04-04 by the [reprex package](http://reprex.tidyverse.org) (v0.2.0).

mvevans89:
@alexpghayes I'm willing to take a stab at this!

[...]

PR #377, opened 2018-06-15
use ldply instead of lapply to tidy aovlist's

mvevans89:
this changes the way tidy is used to tidy aovlist's (such as within-subject error structures) to close issue #299


R/anova_tidiers.R
```diff
@@ -103,22 +103,28 @@ tidy.aov <- function(x, ...) {
...
   }
 
-  # ret <- plyr::ldply(x, tidy, .id = "stratum")
-  ret <- lapply(x, function(a) tidy(stats::anova(a)))
-  ret <- lapply(
-    names(ret),
-    function(a) dplyr::mutate(ret[[a]], stratum = a)
-  )
-  ret <- do.call("rbind", ret)
+  ret <- plyr::ldply(x, tidy, .id = "stratum")
```
alexpghayes: We're looking to move away from `plyr` at the moment. Would you be willing to use `purrr` here instead?

R/anova_tidiers.R
```diff
@@ -103,22 +103,28 @@ tidy.aov <- function(x, ...) {
 
 
 #' @rdname anova_tidiers
+#' 
+#' @importFrom plyr ldply
+#' 
+#' @import dplyr
```
alexpghayes: `dplyr` and `plyr` get imported package-wide in DESCRIPTION, so I believe these aren't necessary.

tests/testthat/test-anova.R
```diff
@@ -0,0 +1,15 @@
+context("anova tidiers")
```
alexpghayes: Regression tests! Awesome!

tests/testthat/test-anova.R
```diff
@@ -0,0 +1,15 @@
...
+
+test_that("tidy.aovlist works"){
+  aovlistfit <- aov(mpg ~ wt + disp + Error(drat), mtcars)
+  td <- suppressWarnings(tidy(aovlistfit))
```
alexpghayes: What kind of warnings are getting suppressed here?

alexpghayes:
Don't worry about the build failure at the moment, BTW, that's my fault and I need to fix it.

R/anova_tidiers.R
mvevans89: Got it. I'll remove them.

tests/testthat/test-anova.R
mvevans89: I checked and there actually are none. Warning suppression was just a hold over from copying the format of the `test-lm.R`. I'll take it out

R/anova_tidiers.R
mvevans89: Works for me. I'll switch it up.

alexpghayes:
Changes look good. Will merge as soon as I sort out the `lavaan` stuff causing the build failure.

PR #382, opened 2018-06-17
Create CODEOWNERS

batpigandme:
* Add @alexpghayes as understudy


alexpghayes:
I get a 404 not found error for https://www.tidyverse.org/development/understudies?

batpigandme:
That's because it doesn't exist. Hadley had me put it there as a placeholder in lieu of having longer text in the CODEOWNERS file itself.

PR #389, opened 2018-06-18
Use unix line endings

alexpghayes:


alexpghayes:
Accidentally amended a commit. Closing and restarting.

Issue #218, opened 2017-04-29
glance.Arima() throws error for time series with `method = "CSS"`


jolars:
`glance()` throws an error when applied to `Arima` objects called with `method = "CSS"`, such as this one:

    fit <- arima(lh, order = c(1, 0, 0), method = "CSS")
    glance(fit)

    Error in `$<-.data.frame`(`*tmp*`, "AIC", value = numeric(0)) : 
       replacement has 0 rows, data has 1

It works fine for the other two methods.

The problem seems to be that `stats::AIC()` and `stats::BIC()` fails for the `Arima` object when the method is sums of squares since there is obviously no log-likelihood to use.

Consider making these calls conditional:

    if (!is.na(ret$lokLik)) {
      ret$AIC <- tryCatch(stats::AIC(x), error = function(e) NULL)
      ret$BIC <- tryCatch(stats::BIC(x), error = function(e) NULL)
    }

josue-rodriguez:
Hi, I'm willing to give this a go!

alexpghayes:
Great! You'll want to start by forking the `broom` repository, and then cloning it to your computer. Once you have it on your local computer, you can start changing the source code (the relevant file is `R/arima_tidiers.R`). You'll want to modify the `glance.Arima` to include the `tryCatch` code about and then you'll want to modify the tests `tests/testthat/test-arima.R` to make sure that example above doesn't error. Once the code passes the test, push it to your repository and make a pull request to broom.

Some resources:
- [Guide to adding new tidiers to `broom`](http://broom.tidyverse.org/articles/adding-tidiers.html)
- [R Packages by Hadley Wickham](http://r-pkgs.had.co.nz/)

It's easy to get information overload, so just drop me a question anytime!

PR #396, opened 2018-06-21
added support for arima objects in glance

josue-rodriguez:
Added glance support for Arima objects when created with method = "CSS" to fix error #218


alexpghayes:
Thanks, this looks great!

Issue #221, opened 2017-05-04
Cannot tidy object of class mlm with conf.int option


shabbybanks:
Here is a MWE:

```r
library(dplyr)
library(broom)
n_o <- 100
set.seed(1234)
datf <- data.frame(x1=rnorm(n_o),x2=runif(n_o),z1=rnorm(n_o),z2=rnorm(n_o)) %>%
  mutate(y1=0.5*x1 + x2 + z1 - z2,y2 = -2*x1 + 0.25*x2 + 3*z1 + z2)
fitm <- lm(cbind(y1,y2) ~ x1 + x2,data=datf)
class(fitm)
```
```
[1] "mlm" "lm"
```
```r
# this runs fine:
broom::tidy(fitm,conf.int=FALSE)
# this does not:
broom::tidy(fitm,conf.int=TRUE)
```
```
Error in data.frame(..., check.names = FALSE) :
  arguments imply differing number of rows: 6, 0

Enter a frame number, or 0 to exit

1: broom::tidy(fitm, conf.int = TRUE)
2: tidy.lm(fitm, conf.int = TRUE)
3: process_lm(ret, x, conf.int = conf.int, conf.level = conf.level, exponentiate = exponentiate)
4: cbind(ret, trans(unrowname(CI)))
5: cbind(deparse.level, ...)
6: data.frame(..., check.names = FALSE)
```

Perhaps a specialized `tidy.mlm` is required.

shabbybanks:
Realizing this is going to be a slog, as `stats::confint` returns empty for objects of class `mlm`, and there is no `confint.mlm`. How unpleasant.

alexpghayes:
If you're interested in writing a `tidy.mlm` method I'd be happy to see a PR! If not, I'll just throw an informative error here, since implementing this is a bit out of scope for `broom`.

shabbybanks:
OK, here's a basic drop-in replacement for `confint.mlm`, though the function API is only appropriate for use in `broom::tidy`:
```R
confint.mlm <- function (object, level = 0.95, ...) {
  cf <- coef(object)
  ncfs <- as.numeric(cf)
  a <- (1 - level)/2
  a <- c(a, 1 - a)
  fac <- qt(a, object$df.residual)
  pct <- stats:::format.perc(a, 3)
  ses <- sqrt(diag(vcov(object)))
  ci <- ncfs + ses %o% fac
  setNames(data.frame(ci),pct)
}

library(dplyr)
library(broom)
[...]

PR #441, opened 2018-07-18
adding mlm tidier

shabbybanks:
Fixing #221 : 
  * adding confidence interval for `mlm` object.
  * quick tidy for `mlm` object inserted into `lm` tidier.
  * adding basic tests.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -31,7 +31,8 @@ Authors@R: c(
     person("Michael", "Kuehn", email = "mkuehn10@gmail.com", role = "ctb"),
     person("Jorge", "Cimentada", email = "cimentadaj@gmail.com", role = "ctb"),
     person("Erle", "Holgersen", email = "erle.holgersen@gmail.com", role = "ctb"),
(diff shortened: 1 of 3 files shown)


shabbybanks:
I noticed that the tests I added do not hit the confidence interval code. I will patch the PR.

shabbybanks:
And now there are these awful CRAN warnings:
```
Unexported object imported by a ':::' call: 'stats:::format.perc'
  See the note in ?`:::` about the use of this operator.
  Including base/recommended package(s):
  'stats'
```
Not sure what your approach is on that one. (TBH, `confint.mlm` should just belong to `stats`, which I would submit a PR for, but I don't think it is on github, and I am not going to deal with `r-dev`.)

shabbybanks:
OK, duh, the unexported function was a one-liner, and I just pasted it in.

alexpghayes:
Fantastic. Will you add yourself as a contributor in `DESCRIPTION` and then I'll merge?

PR #443, opened 2018-07-19
create tidy.rlm method and process_rlm

### LINKED ISSUE #380: `MASS::rlm` tidy output doesn't contain confidence intervals
I think `broom` is currently using `stats::confint` to get confidence intervals for regression estimate from `rlm` objects., but it should instead be using `stats::confint.default`.

``` r

# loading necessary libraries
library(MASS)
library(tidyverse)

# running rlm
x <- MASS::rlm(formula = Sepal.Length ~ Sepal.Width, data = iris)

# getting tidy output
broom::tidy(x = x, conf.int = TRUE)
#>          term   estimate std.error statistic conf.low conf.high
#> 1 (Intercept)  6.6455686 0.4860516 13.672557       NA        NA
#> 2 Sepal.Width -0.2724958 0.1573980 -1.731253       NA        NA

# getting confidence intervals with stats::confint doesn't work
stats::confint(object = x) %>%
  tibble::as_data_frame(x = .)
#> # A tibble: 2 x 2
#>   `2.5 %` `97.5 %`
#>     <dbl>    <dbl>
#> 1      NA       NA
#> 2      NA       NA

# instead using confint.default
stats::confint.default(object = x,
                       level = 0.95) %>%
  tibble::as_data_frame(x = .)
#> # A tibble: 2 x 2
#>   `2.5 %` `97.5 %`
#>     <dbl>    <dbl>
#> 1   5.69    7.60  
#> 2  -0.581   0.0360
```

Created on 2018-06-16 by the [reprex package](http://reprex.tidyverse.org) (v0.2.0).

brunolucian:
This change resolve #380

### DIFF
--- a/R/mass-rlm-tidiers.R
+++ b/R/mass-rlm-tidiers.R
@@ -48,9 +48,23 @@ glance.rlm <- function(x, ...) {
 #' @family rlm tidiers
 #' @seealso [MASS::rlm()]
 #' @export
-tidy.rlm <- function(x, ...) {
-  tidy.lm(x, ...)
+tidy.rlm <- function(x, conf.int = FALSE, conf.level = .95,
+                    exponentiate = FALSE, quick = FALSE, ...) {
+  if (quick) {
+    co <- stats::coef(x)
+    ret <- data.frame(term = names(co), estimate = unname(co),
+                      stringsAsFactors = FALSE)
+    return(process_rlm(ret, x, conf.int = FALSE, exponentiate = exponentiate))
+  }
+  s <- summary(x)
+  ret <- tidy.summary.lm(s)
+  
+  process_rlm(ret, x,
+             conf.int = conf.int, conf.level = conf.level,
+             exponentiate = exponentiate
+  )
 }
+
 
 #' @templateVar class rlm
 #' @template title_desc_augment_lm_wrapper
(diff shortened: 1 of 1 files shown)


brunolucian:
Someone could explain why codecov is unseccessful? @alexpghayes @bschneidr

alexpghayes:
Thanks for this PR!

Rather than reproduce all of `process_lm()` and `tidy.lm()`, I think a more concise approach would be to create `confint.rlm()` function that would get called by [S3 dispatch](https://adv-r.hadley.nz/s3.html). Since we know the `confint.default()` method works, I think including the following line should be enough:

```r
confint.rlm <- confint.default
```

Could you experiment with this approach?

The code coverage errors are because the new code has a lower line cover than the existing code. You'll want to write [tests](http://r-pkgs.had.co.nz/tests.html) to bring this up.

brunolucian:
Sorry, i think i couldn't figured out how to do the way you explain. Could you elaborate more? 
I read about S3 dispatch but i really don't understandig whats going on.

alexpghayes:
Sorry for the long delay here. I'll dig around in here and see what I come up with.

alexpghayes:
Hey @brunolucian. So I ended up fixing this in #483, at https://github.com/tidymodels/broom/pull/483/files#diff-a5be610f6a39294b136d1f8c266ec3afR41. The idea is that the `tidy.rlm` calls `tidy.lm` which calls `confint` on the rlm object. It was using `confint.lm` instead of `confint.rlm` since `rlm` object subclass `lm` objects.

That line creates a `confint` method for `rlm` objects. The result is that `confint.rlm` gets called instead of `confint.lm`. Then we just have to make sure that `confint.rlm` works. But we already know that `confint.default` works for `rlm` objects, so we can do: `confint.rlm <- confint.default` like I did in the file I linked to.

This is preferable to creating a `process_rlm()` function because it avoids a lot of code duplication.

I'm going to close this PR, but want to thank you a ton for diving into this bug and contributing code!

PR #456, opened 2018-07-28
Documentation about multi-dimensional stat::density inputs.

pjpaulpj:
Added an explanation on `stat::density`'s behaviour when the input matrix is a multi-dimensional matrix.


alexpghayes:
Awesome! Thank you for taking this on! Will you add yourself as a contributor in DESCRIPTION and then I'll merge?

Issue #450, opened 2018-07-25
Methods that do not pass strict tests


alexpghayes:
This issue is based on #449, which I assumed will be merged.

A number of tidiers do not yet pass the strict tests. Beginning with the changes in #449, the `strict = TRUE` is the default in tests, so these tidiers can be identified based on the appearance of `strict = FALSE` in the tests file. 

The following tidying methods still do not meet the new `strict = TRUE` specifications:

- [ ] augment.prcomp
- [ ] augment.speedlm
- [ ] augment.rlm
- [ ] augment.rq
- [ ] tidy.map
- [ ] glance.coxph
- [ ] tidy.Mclust
- [ ] tidy.aareg
- [ ] augment.mjoint
- [ ] glance.mjoint
- [ ] augment.smooth.spline
- [ ] tidy.htest
- [ ] glance.htest
- [ ] augment.htest
- [ ] augment.lm -- not sure why this is??
- [ ] augment.clm
- [ ] tidy.survdiff
- [ ] augment.nlrq
- [ ] tidy.polr
- [ ] augment.polr
- [ ] augment.polca
- [ ] augment.glm
- [ ] tidy.cld
- [ ] tidy.kmeans
- [ ] augment.kmeans
- [ ] tidy_emmeans
- [ ] tidy.ridgelm
- [ ] tidy.manova
- [ ] tidy.gam
- [ ] tidy.density
- [ ] tidy.cch
- [ ] tidy.survreg
- [ ] glance.muhaz
- [ ] tidy.spec
- [ ] glance.Arima
- [ ] tidy_xyz_list
- [ ] augment.lmRob
- [ ] augment.glmRob
- [ ] tidy_list_svd / tidy_list_irlba
- [ ] augment.rqs
- [ ] tidy.pyears
- [ ] augment.loess
- [ ] spatial tidiers -- might just punt on this given potential deprecation
- [ ] augment.nls
- [ ] tidy.factanal
- [ ] augment.factanal
- [ ] tidy.btergm

The augment methods in general require the most work. Tidy and glance methods typically require only minor changes and are **beginner friendly** issues. Oftentimes, the only necessary change will be to record some column names in `column_glosssary` in [modeltests](https://github.com/alexpghayes/modeltests)

thisisnic:
I'd love to have a go at fixing this as my first contribution!

ellessenne:
Hi @alexpghayes, I can deal with the methods for the `mjoint` class :wink:

[...]

PR #461, opened 2018-07-31
Update tidier for survival::survdiff

thisisnic:
This change relates to part of #450 

Hope this looks alright - let me know if anything needs updating!

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -32,7 +32,7 @@ Authors@R: c(
     person("Jorge", "Cimentada", email = "cimentadaj@gmail.com", role = "ctb"),
     person("Erle", "Holgersen", email = "erle.holgersen@gmail.com", role = "ctb"),
     person("Karl", "Dunkle Werner", role = "ctb"),
-    person("Steven", "Pav", email = "shabbychef@gmail.com", role="ctb")),
+    person("Steven", "Pav", email = "shabbychef@gmail.com", role="ctb"),
     person("Nic", "Crane", email = "thisisnic@gmail.com", role="ctb"))
 Maintainer: David Robinson <admiral.david@gmail.com>
 Description: Summarizes key information about statistical objects in tidy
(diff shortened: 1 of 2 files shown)


alexpghayes:
Can you write a short description of what tests weren't passing before and how this change fixes them? Since I haven't used the survival tidiers before I have a hard time understanding the change you made.

alexpghayes:
Would you mind writing an example showing the behavior of the tidier before and after those changes? Still trying to wrap my head around this.

thisisnic:
The bit that I edited was inside `tidy.survdiff()`.  Here's an example call to `survdiff()`:
```
x <- survdiff(Surv(time, status) ~ pat.karno + ph.ecog +
                    strata(inst) + strata(sex), data = lung)
```
The code inside `tidy.survdiff()` that I changed originally looked like this:
```
l <- lapply(strsplit(rownames(x$n), ", "), strsplit, "=")
row_list <- lapply(l, function(x)
  structure(
    as.data.frame(lapply(x, "[", 2), stringsAsFactors = FALSE),
    names = sapply(x, "[", 1)
  ))
gvars <- do.call("rbind", row_list)
```
`gvars` is later combined via `bind_cols` with the model results and is what's causing the test failure due to the column names, as show below:
 
```
head(gvars)
# pat.karno ph.ecog
# 1         30       2
# 2         40       2
# 3         50       2
# 4         60       0
# 5         60       1
# 6         60       2
```

So I updated it to just take the exist values in a column called "..." as below:

```  
gvars = data.frame("..." = rownames(x$n), stringsAsFactors = FALSE)
```
So now it's compliant with possible values in `column_glossary`:
```
head(gvars)
# ...
# 1 pat.karno=30, ph.ecog=2
# 2 pat.karno=40, ph.ecog=2
# 3 pat.karno=50, ph.ecog=2
# 4 pat.karno=60, ph.ecog=0
# 5 pat.karno=60, ph.ecog=1
# 6 pat.karno=60, ph.ecog=2
```

However after reading the discussion on the issue page, I'm now unsure whether it'd be better to stick with the previous implementation and set `strict` to `FALSE` in the tests as I don't think my implementation strictly meets the definition of tidy data?

alexpghayes:
Sorry for the delay in responding here.

Yeah, I think the old behavior had a more tidy output. The issue here is that columns in the original data should be allowed in this case. Let's go with `strict = FALSE` here. Would you be willing to document which strict tests fail and why that's okay?

thisisnic:
Yep, I'll get on it now!

Issue #404, opened 2018-06-25
Adding statistic and p-values to `rq` class objects


IndrajeetPatil:
The `tidy.rq` function currently doesn't output either the `statistic` (t-value) or `p.value` for the quantile regression, but this can be easily obtained using `summary.rq()`. It will be nice if this is included in the `tidy` output to be consistent with output from other regression models. The `se = "nid"` part can be absorbed in `...`, so if anybody wants to use a differerent method to compute standard standard errors, they can use this argument.

``` r

# loading libraries and needed data
library(broom)
library(quantreg)
data(engel)

# specifying quantile regression model
rq_model <- quantreg::rq(
  formula = foodexp ~ income,
  tau = 0.5,
  data = engel
)

# getting tidy summary
broom::tidy(x = rq_model)
#>          term   estimate   conf.low   conf.high tau
#> 1 (Intercept) 81.4822474 47.0904023 135.1883939 0.5
#> 2      income  0.5601806  0.4803301   0.6127786 0.5

# getting summary with summary.rq
coef(summary(object = rq_model, se = "nid"))
#>                  Value  Std. Error   t value     Pr(>|t|)
#> (Intercept) 81.4822474 19.25066025  4.232699 3.322875e-05
#> income       0.5601806  0.02827721 19.810319 0.000000e+00
```

Created on 2018-06-25 by the [reprex package](http://reprex.tidyverse.org) (v0.2.0).

ethchr:
I'd be happy to take a crack at this.

alexpghayes:
Awesome! Will follow up with some details tomorrow.

alexpghayes:
You'll want to start by forking the broom repository, and then cloning your fork to your computer. Create a new branch called `tidy-rq-p-values` or something similar. If you are new to git, I recommend [this guide](http://happygitwithr.com/) by Jenny Bryan. Then you'll want to look for `R/quantreg-rq-tidiers.R`. Once you've found the `tidy.rq()` function, you'll need to modify it to include `statistic` and `p.value` columns.

After you've got everything working the way you want, you should [document](http://r-pkgs.had.co.nz/man.html) your work following tidyverse documentation [guidelines](http://style.tidyverse.org/documentation.html).

Then next step will be to [write some tests](http://r-pkgs.had.co.nz/tests.html) to ensure that your functions are working correctly. You'll want to edit the tests in `tests/testthat/test-quantreg-rq.R`.

[...]

PR #466, opened 2018-08-02
Tidy.rq p-values and se default

ethchr:
This is regarding #404. It changes the documentation to indicate that `se = "rank"` cannot return `p.value` and `statistic` columns, and changes the default `se` argument to match that of `summary.rq`, namely, `se = "rank"` for less than 1,000 rows and `se = "nid"` for more than 1,000 rows.


alexpghayes:
Thanks for this pull request! It looks good, I'm going to try to resolve the merge conflicts and see if that gets the build going again.

PR #471, opened 2018-08-07
Multivariate support for tidy.acf

CharlesNaylor:
`acf` and `pacf` support multivariate inputs, in which case the current code winds up providing a data.frame with lots of poorly differentiated lag and acf columns. This commit uses snames to keep acf and lag to single columns with two additional columns describing the time series in 'row' and 'col' positions as they would be displayed in the base acf plot. 

Note that calls to `ccf` will get redundant info in 'row' and 'col'. Arguably, we should provide 'row' and 'col' for univariate acfs, as well, so the output tibble always has the same columns.


CharlesNaylor:
Fixes #471

alexpghayes:
This seems like a useful additional. Can you post a reprex of the new behavior to this thread so I can see what it does?


I'm not sure that `row` and `col` refer to here. Can you think of column names that would be more informative?

It would also be good to:

- Add a test for multivariate acf objects
- Updates NEWs with details about this addition
- Add yourself a contributor in `DESCRIPTION`

CharlesNaylor:
Sure. Here's current behavior:
``` r
library(broom)
x <- diff(log(EuStockMarkets))
str(acf(x, plot=F))
#> List of 6
#>  $ acf   : num [1:27, 1:4, 1:4] 1 -0.000435 -0.026729 -0.010458 0.000307 ...
#>  $ type  : chr "correlation"
#>  $ n.used: int 1859
#>  $ lag   : num [1:27, 1:4, 1:4] 0 0.00385 0.00769 0.01154 0.01538 ...
#>  $ series: chr "x"
#>  $ snames: chr [1:4] "DAX" "SMI" "CAC" "FTSE"
#>  - attr(*, "class")= chr "acf"
str(tidy(acf(x, plot=F)))
#> Classes 'tbl_df', 'tbl' and 'data.frame':    432 obs. of  4 variables:
#>  $ row: Factor w/ 4 levels "DAX","SMI","CAC",..: 1 1 1 1 1 1 1 1 1 1 ...
#>  $ col: Factor w/ 4 levels "DAX","SMI","CAC",..: 1 1 1 1 1 1 1 1 1 1 ...
#>  $ lag: num  0 0.00385 0.00769 0.01154 0.01538 ...
#>  $ acf: num  1 -0.000435 -0.026729 -0.010458 0.000307 ...
```

I named the series indexers `row` and `col` because if you call `acf(x, plot=T)`, it generates a grid of plots of cross-correlations, and the pairs of series on the grid correspond to `row` and `col` as I have them. This way, a similar plot to the base acf plot can be made with `facet_grid(row~col)`.

I added (but didn't push) NEWS and DESCRIPTION. I'll work on a test tomorrow.

CharlesNaylor:
I'm going to switch `col` to `column1` to comply with the modeltests column glossary. Is that acceptable? 

Arguably it should just be `column` if we're sticking with my logic on naming them, in which case it needs to be added to the column glossary. 

Or , if you don't like the `row`, `col` schema, I guess it should be `column1` and `column2` from that list.

I've pushed with it as `column1`.

PR #485, opened 2018-08-13
#450 Update tidier for survival::survdiff

thisisnic:
Update the documentation for survival::survdiff to explain why setting 'strict' to 'FALSE' in the tests is OK, as per conversation here with @alexpghayes - https://github.com/tidymodels/broom/pull/461


alexpghayes:
Awesome, thanks for documenting this!

PR #486, opened 2018-08-13
#450 - update tidy.kmeans and document why strict = FALSE is OK for its tests

thisisnic:
tidy.kmeans uses column names from the input to name columns in the output, and so will not pass tests if strict is set to TRUE.  I have added a comment above the test explaining this.

In addition, the tidier was naming the columns x1...xN (where N is the number of columns in the input).  I have updated it to use the original column names if present, for clarity in the output.  I wasn't sure whether this change would be acceptable given it is beyond the scope of the original issue.

Is this change OK, and if so, do I need to document this elsewhere?  If not, shall I submit a new PR containing just the comments on setting strict to FALSE?


alexpghayes:
I think this is a reasonable change. I would add some documentation that `col.names` should be set to `NULL` to the get variable names `x1`, `x2`, ....

I would also add this change to `NEWS` so that it doesn't surprise anyone.

alexpghayes:
@thisisnic Sorry to leave you hanging for so long on these PRs! Are you still interested in updating the documentation and `NEWS`? If not, I'm happy to do so.

thisisnic:
Sorry @alexpghayes, totally my fault, it's been on my to-do list for ages and kept creeping to the bottom.  I'll look at it now and get it done in the next few days at the latest.  Thanks! :)

thisisnic:
@alexpghayes - I have update and compiled the documentation for tidy.kmeans and put a note in NEWS.md under the "other changes" section.  Hope this all looks OK, if not let me know what I need to update.  Thanks again for your help in guiding me through this process!

PR #490, opened 2018-08-29
Bugfix `tidy.Mclust()` for one-dimensional models. Fixes #489

### LINKED ISSUE #489: Bug for one-dimensional models in `tidy.Mclust()`
When using `tidy.Mclust()` with one-dimensional models, the mean of each component is given in different columns, and all of them reported in the row corresponding to the last component, while the values for the other rows are filled filled with NAs.

If the model is one-dimensional, there should be only one `mean` column with the mean value for each component.

```r
library(dplyr) 
library(mclust)
set.seed(27)

centers <- tibble::tibble(
  cluster = factor(1:3), 
  num_points = c(100, 150, 50),  # number points in each cluster
  x1 = c(5, 0, -3),              # x1 coordinate of cluster center
)

points <- centers %>%
  mutate(
    x1 = purrr::map2(num_points, x1, rnorm)
  ) %>% 
  select(-num_points, -cluster) %>%
  tidyr::unnest(x1)

m <- mclust::Mclust(points)

tidy(m)

#># A tibble: 3 x 7
#>  component  size proportion variance mean.1  mean.2 mean.3
#>      <int> <int>      <dbl>    <dbl>  <dbl>   <dbl>  <dbl>
#>1         1    47      0.167     1.02  NA    NA       NA   
#>2         2   153      0.499     1.02  NA    NA       NA   
#>3         3   100      0.333     1.02  -3.18  0.0961   5.03
```
Expected output would be:
```r
#># A tibble: 3 x 5
#>  component  size proportion variance    mean
#>     <int> <int>      <dbl>    <dbl>   <dbl>
#>1         1    47      0.167     1.02 -3.18  
#>2         2   153      0.499     1.02  0.0961
#>3         3   100      0.333     1.02  5.03 
```

ltobalina:
For one-dimensional models, return the mean for each component in one column.

The previous code was returning the mean of each component in different columns, and all of them reported in the row corresponding to the last component, while the values for the other rows were filled filled with NAs.

Now, if the model is one-dimensional, the mean for each component is stored in a column vector, which is added to the final table `ret` as a new column named `mean`.


alexpghayes:
Thank you for this! I need to update the `modelgenerics` dependency, which should get the AppVeyor test passing, and then I'll take a closer look!

alexpghayes:
The `modelgenerics` update has been taken care of! Would you be willing to write a quick test to confirm that the change you've made fixes the bug? Then you could update `NEWS`, add yourself as a contributor, and I'd merge once the test was passing!

ltobalina:
Hi. Sorry for the delayed response. I've added a couple of tests to check the correct behaviour when the input to `Mclust()` is a one-dimensional vector. The previous fix broke the correct behaviour for when the `G` parameter of `Mclust()` was set to 1. It should work correctly now.

However, if the input data to `Mclust()` is an unnamed vector, `augment.Mclust()` will throw an error. I've added a test case for this, but I haven't found the fix yet.

alexpghayes:
What about inspecting the `data` element of the `Mclust` object and just throwing an informative error?

alexpghayes:
I finished this up in #630. I tried to just push the commits to this PR but my git-fu wasn't good enough, apologies.

ltobalina:
Thanks a lot!

PR #494, opened 2018-09-07
tidy for "mediate" objects

medinali:
broom elements from list returned by mediation::mediate() into a tibble.

**new files**
- man/tidy.mediate.Rd (created with Roxygen)
- R/mediate-tidiers.R -- documented tidy() for method "mediate"

 
Tibble always contains four rows and the number of columns depends on whether conf.int = TRUE

ps. I don't own mediation package.


alexpghayes:
This is fantastic! Thank you!

Issue #480, opened 2018-08-07
Drop the reshape2 dependency


alexpghayes:

PirateGrunt:
I would love to have a go at this. Should I be working on the `master` branch, or is there another? Or should I create a new branch for this?

alexpghayes:
That would be awesome! Definitely work on a new branch!

I'm pretty sure `reshape2` is a dependency because it gets used once to tidy an arbitrary dimension array. I don't think the current tidyverse tools have support for such arrays yet, but would love to have you take a look around and see if there's something you can get to work.

Alternatively, if we could just copy some from `reshape2` so we don't have to depend on the whole package, that would be great.

PirateGrunt:
I've removed the dependency from `tidy.dist` and `list-xyz`. Pull request just updated and being checked.

`reshape2` also appears in the tidiers for `svd`, `polca` and `ks`. In each of those cases, it looks as though it's being used to melt a 3-D array. The fixes for the first two were fairly straightforward. For those remaining, I think `reshape2` may be an appropriate option. Replacing it would basically mean re-writing something that `reshape` already does well, i.e. melting high-dimensional data.

Thoughts?

simonpcouch:
This was addressed in #852.🙂

PR #495, opened 2018-09-08
Remove reshape2 dependency from tidy.dist and add tests.

PirateGrunt:
I am using `tidyr::gather` in the `tidy.dist()` function and have also qualified the use of `filter()` as `dplyr::filter()`. I did not include roxygen comments to import these functions as there were not similar comments for functions like `as_tibble()`. I think the various package dependencies and imports should take care of that.

There is at least one test which failed, but it comes from the file `test-mass-rlm.R`. I pulled from the master branch just before testing.


PirateGrunt:
Meant to reference issue #480 .

alexpghayes:
Awesome! Thanks for taking this on and sorry for taking such a long time to reply!

Would you add yourself as a contributor in `DESCRIPTION` and then I'll merge?

alexpghayes:
Awesome! Thanks a ton!

PR #498, opened 2018-09-10
replace colnames with names in tidy.polr (#497)

### LINKED ISSUE #497: tidy.polr throws bug when conf.int = T
Hello, I found that `tidy.polr` generates a bug when `conf.int = T`. It seems that the `colnames` should be replaced with `names`.

```r
library(tidyverse)
library(broom)
MASS::polr(Species ~ Sepal.Length, data = iris, Hess = T) %>% 
  tidy(exponentiate = T, conf.int = T)
#> Error in `colnames<-`(`*tmp*`, value = c("conf.low", "conf.high")): attempt to set 'colnames' on an object with less than two dimensions
```

haozhu233:
Trying to fix #497

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -44,6 +44,10 @@ The following have all been deprecated in favor of `broom.mixed`:
 - `tidy.lme()`, `glance.lme()`, `augment.lme()`
 - `tidy.stanreg()`, `glance.stanreg()`
 
+## Other changes
+
+- Bug fix for tidy.polr for incorrectly using colnames. (#498)
+
 # broom 0.5.0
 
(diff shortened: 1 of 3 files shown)


alexpghayes:
Thanks for the PR! Would you mind adding a quick test and adding the bugfix to `NEWS` and then I'll merge?

haozhu233:
@alexpghayes I just added NEWs and test cases and I made some changes to how I fixed this. The main problem is that when there is only one numeric x, `stats::confint()` returns a vector instead of a matrix.

alexpghayes:
Thanks for taking this on!

PR #502, opened 2018-09-29
tidy.plm: add argument quick

MatthieuStigler:
`tidy.plm` does not have the standard `quick` argument? I added it explicitly (also in doc, needs to be re-run?) though another approach would be to simply pass the ..., which for now is ignored?

Thanks!


alexpghayes:
Thanks for taking that on!

PR #509, opened 2018-10-10
lfe-tidier: add argument quick to tidy

MatthieuStigler:
Hi

`tidy.felm` did not have the `quick` option, which was very useful in my case. I made an implementation of it, trying to keep code as in other functions. I did local test, but no global ones (difficult to run R CMD check as so many packages required)... I hope it's fine!?

Thanks!

### DIFF
--- a/R/lfe-tidiers.R
+++ b/R/lfe-tidiers.R
@@ -3,6 +3,7 @@
 #'
 #' @param x A `felm` object returned from [lfe::felm()].
 #' @template param_confint
+#' @template param_quick
 #' @param fe Logical indicating whether or not to include estimates of
 #'   fixed effects. Defaults to `FALSE`.
 #' @template param_unused_dots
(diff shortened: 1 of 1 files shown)


alexpghayes:
This looks great! Can you add some tests for `tibble` output when `quick = TRUE`? Or just construct a `tibble` instead of a `data.frame`?

MatthieuStigler:
ok, I did following changes:

- use directly `data_frame`
- added tests with `quick=TRUE` in *test-lfe.R*

I also corrected my own change, where now `quick` takes precedence over `conf.int=TRUE` (i.e. `if (conf.int)` becomes `if (!quick & conf.int)`). This seemed to be the policy in `broom`, am I correct?

Thanks

MatthieuStigler:
ok, I did! It seems now it is not passing the test, but seems for a totally different reason?

PR #511, opened 2018-10-19
Add `tidy` methods for `estimatr` objects

lukesonnet:
This PR adds support for `tidy`ing the four major classes of objects from `estimatr`: `lm_robust`, `iv_robust`, `difference_in_means`, and `horvitz_thompson`. This PR does not add `augment` or `glance` methods.

The `estimatr` package had a hacked version of `tidy` before to avoid collisions with broom and to avoid the dependency. We now would rather PR broom and DEPEND on broom.

Because of this hack, there may be some testing weirdness with `estimatr`, but so far it seems like `broom` passes tests with either the CRAN version of `estimatr` or the version of `estimatr` we'll push to CRAN once you update `broom` on CRAN (which can be found here, soon to be merged in: https://github.com/DeclareDesign/estimatr/pull/262).

I've followed your contributors guide as best as I could, please let me know if you have any questions.


lukesonnet:
Both errors seem to be on the CI side (rgeos failed installation and rate limiting downloads).

PR #514, opened 2018-10-22
lfe-tidyr: allow for multiple response

MatthieuStigler:
Hi

Like lm(), `felm()` can compute multi-responses models, using `lfe::felm(v1 + v2 ~ v3 , df)`. `tidy()` will fail as `summary()` on such an output returns: *Error in summary.felm(x) : Please specify lhs=[one of v1,v2]* 

This pull request calls summary() for each response, then binds it so that tidy() work on it. For `glance()` and `augment()`, it follows the behaviour of `glance.mlm`, that will not return anything. I am not sure whether this is a long-term strategy of the package, or not? There are clearly cases where allowing for multiple responses at that level would be great!

Thanks!

### DIFF
--- a/R/lfe-tidiers.R
+++ b/R/lfe-tidiers.R
@@ -148,7 +148,7 @@ augment.felm <- function(x, data = model.frame(x), ...) {
   
   if(has_multi_response) {
     stop(
-      "Glance does not support linear models with multiple responses.",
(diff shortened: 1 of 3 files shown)


alexpghayes:
Awesome. Will take a look over the upcoming weekend!

alexpghayes:
Looks good, thanks!

PR #519, opened 2018-10-25
save object as tibble and increase coverage

medinali:
- this changes columns' type from chr to dbl
- increase coverage of tidy.mediate() to 100


alexpghayes:
Thanks for the PR!

PR #530, opened 2018-11-17
Add standard error to tidy_optim

### LINKED ISSUE #529: Add Standard Error for `optim()`, if Hessian is available
It would be helpful if `tidy_optim()` supported summarizing the Hessian, if present.

``` r
library(broom)

tidy_optim <- function(x, ...) {
  if (is.null(names(x$par))) {
    names(x$par) <- paste0("parameter", seq_along(x$par))
  }
  ret <- tibble(parameter = names(x$par), value = unname(x$par))
  if ("hessian" %in% names(x)) {
    ret$std.error <- sqrt(diag(solve(x$hessian)))
  }
  ret
}

r <- function(x) (3 - x[1])^2 + 4 * (x[2] - x[1]^2)^2

res <- optim(1:2, r)
tidy_optim(res)
#> Error in tibble(parameter = names(x$par), value = unname(x$par)): could not find function "tibble"

resH <- optim(1:2, r, hessian=TRUE)
tidy_optim(resH)
#> Error in tibble(parameter = names(x$par), value = unname(x$par)): could not find function "tibble"
```

Created on 2018-11-16 by the [reprex package](http://reprex.tidyverse.org) (v0.2.0).

(As mentioned in #505)

billdenney:
Fix #529


alexpghayes:
I'm a fan. Can you a quick test and update the documentation?

alexpghayes:
No need to do anything to base tests, those are already there. I would just add a test to https://github.com/tidymodels/broom/blob/master/tests/testthat/test-list-optim.R#L19 that makes sure the `std.error` column is present when applicable, and isn't when it shouldn't be. This might surprising to users, so documenting it carefully will be important!

Does that help?

alexpghayes:
Thanks!

PR #534, opened 2018-11-21
Adapt to dev tibble

jennybc:
I'm helping @krlmlr get out ahead of revdep breakages associated with an upcoming release of tibble. I selected this package, because it's one of the revdeps with considerable download activity.

tibble is getting stricter about tibbles with unnamed columns or duplicate column names. There are, of course, ways to explicitly allow that or request name repair, but that code would depend on a specific version of tibble.

This PR makes broom work with the current CRAN version of tibble (v1.4.2) and with the coming release.

Main theme: in order to write code that does not tickle the stricter behaviour around names, i.e. that is robust to tibble version, apply column names before applying `as_tibble()`, vs. after.


alexpghayes:
Okay there's now a branch https://github.com/tidymodels/broom/tree/0.5.0_patches. 

If there's a cleaner git solution please let me know, but otherwise let's roll with this.

jennybc:
Sounds reasonable. What version number do you plan for the CRAN release? Perhaps that's what the branch should be called and the first commit there should make it so in DESCRIPTION.

I'm happy to close this PR and make same against that branch when you are ready.

alexpghayes:
Okay we're live at https://github.com/tidymodels/broom/tree/0.5.1.

jennybc:
Closing. We're taking this up over in https://github.com/tidymodels/broom/pull/536 now.

PR #536, opened 2018-11-22
Adapt to dev tibble

jennybc:
Resuming https://github.com/tidymodels/broom/pull/534 here.

sp-tidiers main theme: apply column names before coercing to tibble

lavaan-tiders: use `enframe()`, which is unambiguously requesting a column, instead of `as_tibble()` on a vector

base tidiers: I gather these are on the path to deprecation / removal, since they are basically just `as_tibble()`. So I altered your tests and examples enough that you aren't testing our edge cases. I also coerce `ftable` objects to `table` so we only have to worry about one thing.

### DIFF
--- a/R/base-tidiers.R
+++ b/R/base-tidiers.R
@@ -14,7 +14,7 @@
 #' 
 #' @examples
 #'
-#' tab <- with(airquality, table(cut(Temp, quantile(Temp)), Month))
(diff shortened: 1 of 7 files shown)


jennybc:
I rebased this and cherry-picked the AppVeyor fix. Maybe now we can get good data from CI.

PR #537, opened 2018-11-22
Install Rtools and cache the library on appveyor

jennybc:
I think this might clear up some of your appveyor failures and possibly speed things up. With this many dependencies, it seems likely that *something* is going to need compilation.

If this seems good, you might want to port to the 0.5.1 branch as well.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -10,6 +10,13 @@ init:
 install:
   ps: Bootstrap
 
+environment:
+  global:
+    USE_RTOOLS: true
+
+cache:
+  - C:\RLibrary
+
 # Adapt as necessary starting from here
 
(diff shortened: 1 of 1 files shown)


alexpghayes:
Awesome, thanks! At some point all the AppVeyor builds starting running into installation issues and crashing, but before that they were taking about 10 minutes. Not sure why the package installation issues popped up. Anyway the whole CI setup scares me a bit so thanks for this!

jennybc:
This seems to improve the AppVeyor situation. I think some similar improvements may be possible on travis as well. I might do that in a separate PR.

alexpghayes:
Awesome, thanks!

PR #538, opened 2018-11-22
Remove sudo:required

jennybc:
An experiment. Let's see what happens. I think you may be able to remove more things from `travis.yml`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,6 @@ r:
   - release
   - devel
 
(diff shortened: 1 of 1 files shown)


jennybc:
Hmm, perhaps you cannot do this without making other changes. You might want to ask Jim H. to take a look. He is good at spotting cruft in `.travis.yml` that makes things slower than they have to be.

Issue #543, opened 2018-11-28
Set default arguments for methods to pass strict tests


jmuhlenkamp:
I intend to open a PR for this issue.

There are a number of tidier methods that do not pass strict tests because default arguments required for strict tests have not been set.  For example.

    check_arguments(augment.factanal)
    ## Error: length(no_defaults) == 0 isn't true.
    ## Arguments data to `augment.factanal` must have default values. Use `NULL` to indicate missingness if necessary.

***Change 1: Add default arguments***

In order to pass `check_argument` tests, default arguments will be added for the following methods and arguments:
- `augment.betareg`: `type.predict`, `type.residuals`
- `augment.factanal`: `data`
- `augment.kmeans`:  `data`
- `augment.prcomp`: `newdata`
- `augment.rqs`: `newdata`

***Change 2: Remove unnecessary strict = FALSE***

Furthermore, there are 4 `check_argument` tests where `strict = FALSE` where as `strict = TRUE` passes (see below).  These tests will be updated to remove `strict = FALSE`.

    check_arguments(augment.ivreg)
    check_arguments(tidy.lsmobj)
    check_arguments(augment.Mclust)
    check_arguments(augment.clm)

***Change 3: Update internal data to remove warning***

Finally, `augment.factanal` currently throws a warning if `data` is specified.  The internal data structure will be updated to prevent coercion.

    library(broom)
    fit_reg <- factanal(mtcars, 3, scores = "regression")
    au_reg <- augment(fit_reg)
    au_reg2 <- augment(fit_reg, data = mtcars)
    ## Warning message:
    ## Column `.rownames` joining factor and character vector, coercing into character vector 

I do not expect/intend of these changes to be breaking.

PR #544, opened 2018-11-28
Meet specification more fully (i.e. enable more strict tests)

jmuhlenkamp:
This is a pull request for https://github.com/tidymodels/broom/issues/543 (also general strict tests issue https://github.com/tidymodels/broom/issues/450).

Would appreciate some feedback on how I am setting the default arguments for `augment.betareg` as it feels a bit awkward (see below for snippet). 

    augment.betareg <- function(x, data = model.frame(x), newdata = NULL,
                                type.predict, type.residuals, ...) {
                                type.predict = eval(formals(betareg:::predict.betareg)$type),
                                type.residuals = eval(formals(betareg:::residuals.betareg)$type),
                                ...) {


alexpghayes:
This is awesome! I'm swamped with school at the moment but will try to take a look soon!

alexpghayes:
This is really really awesome! Thank you!

For `augment.betareg()` I would either:

- use `c("response", "link", "precision", "variance", "quantile")` for `type.predict`
- use `c("sweighted2", "deviance", "pearson", "response", "weighted", "sweighted")` for `type.predict`

The concern is then what happens if `betareg` changes these options are no longer correct. To address this you could write some tests to make sure all these options work.

Alternatively, you could use `NULL` for the default and then use something like:

```r
type.predict <- match.arg(type.predict, choices = formals(betareg:::predict.betareg)$type)
```

in the body of the function. I think the first option is preferable, although I'm not totally sure.

jmuhlenkamp:
When I get a chance, I'll implement the first option, add a basic test for each options, update the NEWS, and then this PR should be ready to be merged.  Will let you know.

jmuhlenkamp:
This should now be ready to be merged.  Let me know if there is anything I need to do.

jmuhlenkamp:
Thanks!  From a quick look, all the comments make sense and should be pretty minor changes.  Will try to push updates in the next few days.  Will let you know.

alexpghayes:
I know it's been forever, and I'm sorry! This is on my TODO list again!

R/betareg-tidiers.R
```diff
@@ -70,7 +70,9 @@ tidy.betareg <- function(x, conf.int = FALSE, conf.level = .95, ...) {
 #' @seealso [augment()], [betareg::betareg()]
 #' @export
 augment.betareg <- function(x, data = model.frame(x), newdata = NULL,
-                            type.predict, type.residuals, ...) {
+                            type.predict = c("response", "link", "precision", "variance", "quantile"),
```
alexpghayes: In addition to adding these arguments, we should use `arg_match()` to validate them. That is, you'll want to add a line:

```r
type.predict <- arg_match(type.predict)
```

in the first few lines of `augment.betareg()`.

R/stats-factanal-tidiers.R
```diff
@@ -75,8 +75,8 @@ augment.factanal <- function(x, data, ...) {
   }
 
   # Place relevant values into a tidy data frame
-  tidy_df <- data.frame(.rownames = rownames(scores), data.frame(scores)) %>%
-    as_tibble()
+  tidy_df <- bind_cols(tibble(.rownames = rownames(scores)),
```
alexpghayes: You might be able to just do `tibble(.rownames = rownames(scores), scores)`?

R/stats-kmeans-tidiers.R
```diff
@@ -42,7 +42,10 @@ tidy.kmeans <- function(x, col.names = colnames(x$centers), ...) {
 #' @export
 #' @seealso [augment()], [stats::kmeans()]
 #' @family kmeans tidiers
-augment.kmeans <- function(x, data, ...) {
+augment.kmeans <- function(x, data = NULL, ...) {
+  if (is.null(data)) {
```
alexpghayes: In cases like this, where the argument is required, it's actually better to not have a default argument so the user gets an error and knows they have to specify the argument.

alexpghayes:
Looking good! Added some comments!

R/stats-factanal-tidiers.R
jmuhlenkamp: Your suggestion gives a different result (scores is all jammed into a single column).  How about this?

`tidy_df3 <- cbind(tibble(.rownames = rownames(scores)), scores)`

Here is a simple reprex that shows results of this line of code:

```
library(magrittr)
library(tibble)
x <- fit_reg
data = mtcars
scores <- x$scores
tidy_df <- data.frame(.rownames = rownames(scores), data.frame(scores)) %>% 
  as_tibble()
tidy_df2 <- tibble(.rownames = rownames(scores), scores)
tidy_df3 <- cbind(tibble(.rownames = rownames(scores)), scores)

str(tidy_df)
str(tidy_df2)
str(tidy_df3)
```
Here are the data.frames of:
1) what  I currently have in the code
2) your suggestion
3) my new suggestion
```
> str(tidy_df)
Classes ‘tbl_df’, ‘tbl’ and 'data.frame':	32 obs. of  4 variables:
 $ .rownames: Factor w/ 32 levels "AMC Javelin",..: 18 19 5 13 14 31 7 21 20 22 ...
 $ Factor1  : num  0.847 0.722 0.686 -0.866 -0.893 ...
 $ Factor2  : num  0.672 0.384 -0.592 -0.673 0.862 ...
 $ Factor3  : num  -0.2783 0.0246 -0.5644 -0.7666 -1.015 ...
> str(tidy_df2)
Classes ‘tbl_df’, ‘tbl’ and 'data.frame':	32 obs. of  2 variables:
 $ .rownames: chr  "Mazda RX4" "Mazda RX4 Wag" "Datsun 710" "Hornet 4 Drive" ...
 $ scores   : num [1:32, 1:3] 0.847 0.722 0.686 -0.866 -0.893 ...
  ..- attr(*, "dimnames")=List of 2
  .. ..$ : NULL
  .. ..$ : chr  "Factor1" "Factor2" "Factor3"
> str(tidy_df3)
'data.frame':	32 obs. of  4 variables:
 $ .rownames: chr  "Mazda RX4" "Mazda RX4 Wag" "Datsun 710" "Hornet 4 Drive" ...
 $ Factor1  : num  0.847 0.722 0.686 -0.866 -0.893 ...
 $ Factor2  : num  0.672 0.384 -0.592 -0.673 0.862 ...
 $ Factor3  : num  -0.2783 0.0246 -0.5644 -0.7666 -1.015 ...
```

R/betareg-tidiers.R
jmuhlenkamp: This makes sense, will add this once we are aligned on resolution(s) for your other comments.

R/stats-kmeans-tidiers.R
jmuhlenkamp: I'm confused by your comment here.  According to https://github.com/tidymodels/broom/issues/450 

"A number of tidiers do not yet pass the strict tests. Beginning with the changes in #449, the strict = TRUE is the default in tests, so these tidiers can be identified based on the appearance of strict = FALSE in the tests file."

Thus, I thought the overall request was to make revisions such that the tests could run with strict=TRUE.  i.e. `check_arguments(broom:::augment.kmeans, strict = TRUE)`, which currently throws the error message `Arguments data to `augment.kmeans` must have default values. Use `NULL` to indicate missingness if necessary.`. Suggesting that the default argument here should be NULL rather than not having a default..

Also, isn't it better to be explicit here and throw an error saying that data is a required argument  (i.e. `stop("Must specify `data` argument.", call. = FALSE)`)
rather than letting whatever error will surface downstream get fed back to the user?

R/stats-factanal-tidiers.R
alexpghayes: Okay let's keep what's there then.

R/stats-kmeans-tidiers.R
alexpghayes: In this particular case we want the default error for consistency. The `kmeans` tidiers are a special case: the `data` is required, whereas with other tidiers it's optional. So if the user doesn't specify `data` they should get the standard R warning for: please give this function that argument it needs that you forgot.

alexpghayes:
Finally followed up!

R/stats-kmeans-tidiers.R
jmuhlenkamp: Ok, thanks for the explanation.  To close out this discussion, this means the end user message will be:

`Error in rownames(x) : argument "data" is missing, with no default`

I do think there is an opportunity to make the error message more user-friendly such as:

`Must specify "data" argument for augment.kmeans.`

but will defer to you and remove my change and add a comment so future developers are aware not to attempt a similar change.

For convenience, here's a reprex for the current error message.

```
set.seed(2)
x <- rbind(
  matrix(rnorm(100, sd = 0.3), ncol = 2),
  matrix(rnorm(100, mean = 1, sd = 0.3), ncol = 2)
)
fit <- kmeans(x, 2)
augment(fit)
## Error in rownames(x) : argument "data" is missing, with no default 
```

jmuhlenkamp:
Thanks!  I added one comment to close out the `kmeans` discussion, but I should be good to make the requested change(s) now. :-)

R/stats-kmeans-tidiers.R
alexpghayes: If you feel strongly that the original error message is better I'm happy to go with that!

jmuhlenkamp: Cool, thanks for the flexibility!  I can still remember the days before I knew how to dig into R source code and how unfriendly some R error messages can be.

R/betareg-tidiers.R
jmuhlenkamp: This has been added.  Resolving conversation.

R/stats-kmeans-tidiers.R
jmuhlenkamp: Just pushed an update that updates my original error message from `Must specify "data" argument.` to `Must specify "data" argument for augment.kmeans.`

Resolving conversation.

jmuhlenkamp:
@alexpghayes added two commits that close out the open review items.  If this PR is good now, I can work to resolve the merge conflicts and/or merge master into my branch.  Let me know.  Also, I'm not too familiar with the CI build fails, I believe some previous changes from master that I merged into my branch are causing this, but I am not sure.

Let me know how you'd like to me proceed to keep this moving along.  Thanks!

alexpghayes:
Sorry that I've let this sit so long. If you resolve merge conflicts and ping me I'll make sure to review quickly.

Issue #500, opened 2018-09-12
Warning "Multiple parameters; naming those columns num.df, denom.df" expected?


IndrajeetPatil:
Just making sure that this warning is expected in this context? 

``` r
stats::oneway.test(formula = Sepal.Length ~ Species, data = iris)
#> 
#>  One-way analysis of means (not assuming equal variances)
#> 
#> data:  Sepal.Length and Species
#> F = 138.91, num df = 2.000, denom df = 92.211, p-value < 2.2e-16

broom::tidy(stats::oneway.test(formula = Sepal.Length ~ Species, data = iris))
#> Multiple parameters; naming those columns num.df, denom.df
#> # A tibble: 1 x 5
#>   `num df` `denom df` statistic  p.value method                           
#>      <dbl>      <dbl>     <dbl>    <dbl> <chr>                            
#> 1        2       92.2      139. 1.51e-28 One-way analysis of means (not a~
broom::tidy(stats::aov(formula = Sepal.Length ~ Species, data = iris))
#> # A tibble: 2 x 6
#>   term         df sumsq meansq statistic   p.value
#>   <chr>     <dbl> <dbl>  <dbl>     <dbl>     <dbl>
#> 1 Species       2  63.2 31.6        119.  1.67e-31
#> 2 Residuals   147  39.0  0.265       NA  NA
```

Created on 2018-09-12 by the [reprex package](https://reprex.tidyverse.org) (v0.2.0.9000).

IndrajeetPatil:
I have been wondering whether to be internally consistent among `broom` tidiers, the `num.df` and `denom.df` columns here should be renamed to `parameter1` and `parameter2`. 

All other tidiers use the column name `parameter` for degrees of freedom (`df`).

IndrajeetPatil:
@simonpcouch Should this be closed in favor of https://github.com/tidymodels/broom/issues/951?

simonpcouch:
Yeah, thanks @IndrajeetPatil! That renaming message has been clarified.

PR #549, opened 2018-12-07
Make column names of output of htest-oneway tidier syntactically valid 

karissawhiting:
@alexpghayes - In `R/stats-htest-tidiers.R`  there is a message indicating that columns will be renamed to `num.df` and `denom.df` `if (length(x$parameter) > 1)` is `TRUE`, however this renaming doesn't actually occur as far as I can tell (although I may be missing something). 

Here is the relating section of code in the current version:
``` 
x <- oneway.test(extra ~ group, data = sleep)
ret <- x[c("estimate", "statistic", "p.value", "parameter")]

# parameter may have multiple values as well, such as oneway.test
if (length(x$parameter) > 1) {
  ret$parameter <- NULL
  if (is.null(names(x$parameter))) {
    warning("Multiple unnamed parameters in hypothesis test; dropping them")
  } else {
    message(
      "Multiple parameters; naming those columns ",
      paste(make.names(names(x$parameter)), collapse = ", ")
    )
    ret <- append(ret, x$parameter, after = 1)
  }
}
#> Multiple parameters; naming those columns num.df, denom.df
names(ret)
#> [1] NA          "num df"    "denom df"  "statistic" "p.value"
```
I propose the addition of `names(x$parameter) <- make.names(names(x$parameter))`  to make these names syntactically valid:

``` 
x <- oneway.test(extra ~ group, data = sleep)
ret <- x[c("estimate", "statistic", "p.value", "parameter")]

# parameter may have multiple values as well, such as oneway.test
if (length(x$parameter) > 1) {
  ret$parameter <- NULL
  if (is.null(names(x$parameter))) {
    warning("Multiple unnamed parameters in hypothesis test; dropping them")
  } else {
    message(
      "Multiple parameters; naming those columns ",
      paste(make.names(names(x$parameter)), collapse = ", ")
    )
    names(x$parameter) <- make.names(names(x$parameter))
    ret <- append(ret, x$parameter, after = 1)
  }
}
#> Multiple parameters; naming those columns num.df, denom.df
names(ret)
#> [1] NA          "num.df"    "denom.df"  "statistic" "p.value"
```
This is meant as the first step to fixing the oneway-htest tidier and glance functions (relating to issue #450) so they can pass the strict tests in `tests/testthat/test-stats-htest.R`. The output of `stats::oneway.test` currently has `num df` and `denom df` as parameters, but these are not in the column glossary.

If `denom.df` and `num.df` are added to the column glossary for `tidy` and `glance` methods after this change, these should pass the strict test (sidenote: that `num.df`is already in the column glossary for the `tidy` method).

To finish this strict test issue fix, I propose these follow up steps (after this PR is merged).

1. Make a PR on [modeltests](https://github.com/alexpghayes/modeltests) with a change on the `column_glossary.rda` file binding the following additions:

```
 additions <- tribble(
  ~method, ~column, ~description, 
  "tidy", "denom.df", "Denominator degrees of freedom.", 
  "glance", "num.df", "Degrees of freedom.",
  "glance", "denom.df", "Denominator degrees of freedom.") 
```

 2. After that change is merged, make an additional PR here changing `strict = TRUE` in the `check_tidy_outputs` and `check_glance_outputs` functions for the oneway tests of `tests/testthat/test-stats-htest.R`. 

Please let me know if you agree with this plan. Thanks!


IndrajeetPatil:
I think this PR solves https://github.com/tidymodels/broom/issues/500

karissawhiting:
@alexpghayes - Let me know if you had something else in mind for tests regarding `num df` and `denom df` in the output. Thanks!

tests/testthat/test-stats-htest.R
```diff
@@ -19,6 +19,8 @@ test_that("tidy.htest/oneway.test", {
   gl <- glance(ot)
   
   check_tidy_output(td, strict = FALSE)
+  expect_true(!("num df" %in% colnames(td)))
```
alexpghayes:
```suggestion
  expect_false("num df" %in% colnames(td))
```

tests/testthat/test-stats-htest.R
```diff
@@ -19,6 +19,8 @@ test_that("tidy.htest/oneway.test", {
...
+  expect_true(!("denom df" %in% colnames(td)))
```
alexpghayes:
```suggestion
  expect_false("denom df" %in% colnames(td))
```

tests/testthat/test-stats-htest.R
alexpghayes: Would minorly improve readability since you don't have to mentally parse the negative

alexpghayes:
Looks good! Made a small suggestion, but we're close to merging!

alexpghayes:
Awesome! Thanks for contributing!

PR #550, opened 2018-12-09
fix ordinal+scale bug

bbolker:
current version of the `ordinal` tidiers uses a naive (and wrong) approach to distinguish parameters that determine the intercept (`alpha`); effects of parameters on the location (`beta`); effects of parameters on the scale (`zeta`).  It fails if there are identical terms in the location and scale models, which is typical.

Might also be worth considering a **breaking** change to label the coefficient types "intercept", "location", and "scale" instead of "alpha", "beta", and "zeta" ...


alexpghayes:
Restarted the build. Would you open an issue about renaming the coefficient types?

alexpghayes:
Thank you!

PR #551, opened 2018-12-11
preserve aspect ratio of hex logo

tyluRp:
The hex logo was bugging me a bit 😅 not sure why the diff is so large though...


alexpghayes:
Thanks, nice catch!

PR #555, opened 2018-12-16
Fix tidy.survreg

thisisnic:
The test for `tidy.survreg` was failing the  `check_arguments` test because `tidy.survreg` had a `conf.level` argument but no `conf.int` argument.  I made the following changes:

* Updated `tidy.survreg` to also include a `conf.int` argument which is set to FALSE by default (note: previously, confidence intervals were always reported).
* Added this to NEWS as this will be a potentially breaking change.
* Updated the documentation for `tidy.survreg` to include this new parameter.
* Updated the examples for `tidy.survreg` as the coefficient plot needs a `tidy.survreg` object which has been created with `conf.int` set to TRUE.
* Updated the tests for `tidy.survreg` to run `check_arguments` with strict set to TRUE (the default) and also modified the testing of the values to test both when conf.int is set to TRUE or set to FALSE.

Happy to make further changes if any of this need altering though!

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -19,6 +19,7 @@ in the augment method for the chi sq test, .residuals column was renamed to .res
 
 - augment
 - tidy.lsmobj gains a `conf.int` argument.
+- tidy.survreg gains a `conf.int`.
 - Add `tidy.regsubsets()` for best subsets linear regression from the `leaps` package
 - All `conf.int` arguments now default to `FALSE`.
 - All `conf.level` arguments now default to `TRUE`.
--- a/NEWS.md
+++ b/NEWS.md
@@ -19,7 +19,7 @@ in the augment method for the chi sq test, .residuals column was renamed to .res
 
 - augment
 - tidy.lsmobj gains a `conf.int` argument.
-- tidy.survreg gains a `conf.int`.
+- tidy.survreg gains a `conf.int` argument.
 - Add `tidy.regsubsets()` for best subsets linear regression from the `leaps` package
 - All `conf.int` arguments now default to `FALSE`.
 - All `conf.level` arguments now default to `TRUE`.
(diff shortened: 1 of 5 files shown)


NEWS.md
```diff
@@ -19,6 +19,7 @@ in the augment method for the chi sq test, .residuals column was renamed to .res
 
 - augment
 - tidy.lsmobj gains a `conf.int` argument.
+- tidy.survreg gains a `conf.int` argument.
```
alexpghayes: Would you add a note here that previously the confidence interval was included by default but the new behavior does not include a confidence interval by default?

R/survival-survreg-tidiers.R
```diff
@@ -2,7 +2,8 @@
 #' @template title_desc_tidy
 #'
 #' @param x An `survreg` object returned from [survival::survreg()].
-#' @param conf.level confidence level for CI
+#' @param conf.int Logical indicating whether or not to include a confidence interval in the tidied output. Defaults to FALSE.
```
alexpghayes: For consistency throughout the documentation it would nice to use `@template param_confint` here, which will pull in the boilerplate from `man-roxygen/param_confint.R` about how the `conf.int` and `conf.level` arguments are used.

thisisnic: Will do, and thanks, I learned something new today! 😁

NEWS.md
thisisnic: Done - let me know if the phrasing is clear enough or needs reworking.

R/survival-survreg-tidiers.R
```diff
@@ -2,7 +2,8 @@
 #' @template title_desc_tidy
 #'
 #' @param x An `survreg` object returned from [survival::survreg()].
-#' @param conf.level confidence level for CI
+#' @template param_conflevel
```
alexpghayes:
```suggestion
```

Both `conf.int` and `conf.level` are documented in the `param_confint` template (https://github.com/tidymodels/broom/blob/master/man-roxygen/param_confint.R#L3), so we can get away with just `#' @template param_confint` here.

Issue #560, opened 2018-12-20
glance method for biglm doesn't contain residual degrees of freedom


IndrajeetPatil:
Unlike other regression objects, when present, glance method for `biglm` doesn't contain a column for `df.residual` although it is present. Will be nice to have it for consistency.
``` r

# set up
set.seed(123)
library(biglm)
#> Loading required package: DBI
library(broom)

# model
bfit <- biglm::biglm(
  formula = scale(mpg) ~ scale(wt) + scale(disp),
  data = mtcars
)

# tidy output
broom::tidy(bfit, conf.int = TRUE)
#> # A tibble: 3 x 6
#>   term         estimate std.error p.value conf.low conf.high
#>   <chr>           <dbl>     <dbl>   <dbl>    <dbl>     <dbl>
#> 1 (Intercept)  4.68e-17    0.0855 1.000     -0.168   0.168  
#> 2 scale(wt)   -5.44e- 1    0.189  0.00400   -0.914  -0.174  
#> 3 scale(disp) -3.64e- 1    0.189  0.0538    -0.735   0.00593

# glance output
broom::glance(bfit)
#> # A tibble: 1 x 3
#>   r.squared   AIC deviance
#>       <dbl> <dbl>    <dbl>
#> 1     0.781  12.8     6.79

# residual degree of freedom
bfit$df.resid
#> [1] 29
```

<sup>Created on 2018-12-20 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>

<details>

<summary>Session info</summary>

``` r
devtools::session_info()
#> - Session info ----------------------------------------------------------
#>  setting  value                                             
#>  version  R Under development (unstable) (2018-11-30 r75724)
#>  os       Windows 10 x64                                    
#>  system   x86_64, mingw32                                   
#>  ui       RTerm                                             
#>  language (EN)                                              
#>  collate  English_United States.1252                        
#>  ctype    English_United States.1252                        
#>  tz       Asia/Calcutta                                     
#>  date     2018-12-20                                        
#> 
#> - Packages --------------------------------------------------------------
#>  package     * version     date       lib
#>  assertthat    0.2.0       2017-04-11 [1]
#>  backports     1.1.3       2018-12-14 [1]
#>  biglm       * 0.9-1       2013-05-16 [1]
#>  bindr         0.1.1       2018-03-13 [1]
#>  bindrcpp      0.2.2       2018-03-29 [1]
[...]

PR #573, opened 2019-01-06
Feature: have glance.biglm return df.residual

mattle24:
Addressing https://github.com/tidymodels/broom/issues/560


alexpghayes:
Thanks!

Issue #571, opened 2019-01-05
glance method not working for `lavaan` objects


IndrajeetPatil:
``` r
# setup
library(lavaan)
#> This is lavaan 0.6-3
#> lavaan is BETA software! Please report any bugs.
set.seed(123)

# model
cfa.fit <- cfa(
  'F =~ x1 + x2 + x3 + x4 + x5',
  data = HolzingerSwineford1939, group = "school"
)

# glance
broom::glance(cfa.fit)
#> Error in .f(.x[[i]], ...): object 'rmsea.ci.upper' not found
```

<sup>Created on 2019-01-05 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>

<details>

<summary>Session info</summary>

``` r
devtools::session_info()
#> - Session info ----------------------------------------------------------
#>  setting  value                                             
#>  version  R Under development (unstable) (2018-11-30 r75724)
#>  os       Windows 10 x64                                    
#>  system   x86_64, mingw32                                   
#>  ui       RTerm                                             
#>  language (EN)                                              
#>  collate  English_United States.1252                        
#>  ctype    English_United States.1252                        
#>  tz       Asia/Calcutta                                     
#>  date     2019-01-05                                        
#> 
#> - Packages --------------------------------------------------------------
#>  package     * version     date       lib
#>  assertthat    0.2.0       2017-04-11 [1]
#>  backports     1.1.3       2018-12-14 [1]
#>  bindr         0.1.1       2018-03-13 [1]
#>  bindrcpp      0.2.2       2018-03-29 [1]
#>  broom         0.5.1.9000  2018-12-24 [1]
#>  callr         3.1.1       2018-12-21 [1]
#>  cli           1.0.1.9000  2018-10-30 [1]
#>  crayon        1.3.4       2017-09-16 [1]
#>  desc          1.2.0       2018-10-30 [1]
#>  devtools      2.0.1       2018-10-26 [1]
#>  digest        0.6.18      2018-10-10 [1]
#>  dplyr         0.7.8       2018-11-10 [1]
#>  evaluate      0.12        2018-10-09 [1]
#>  fs            1.2.6       2018-08-23 [1]
#>  generics      0.0.2       2018-11-29 [1]
#>  glue          1.3.0       2018-07-17 [1]
#>  highr         0.7         2018-06-09 [1]
#>  htmltools     0.3.6       2017-04-28 [1]
#>  knitr         1.21        2018-12-10 [1]
[...]

PR #577, opened 2019-01-08
Patch lavaan glance bug

mattle24:
`glance.lavaan()` is returning an error currently. This appears to be because `as_tibble()` is not passing on the names of the lavaan fit measures to the created tibble since tibble 2.0.0 defaults to NULL row names. 

Addressing https://github.com/tidymodels/broom/issues/571


alexpghayes:
Thanks!

Issue #562, opened 2018-12-24
glance method not working for `geeglm` objects


IndrajeetPatil:
``` r

# creating a dataframe
set.seed(123)
library(geepack)
data(state)
ds <- data.frame(state.region, state.x77)

# model
geefit <-
  geepack::geeglm(
    formula = Income ~ Frost + Murder,
    id = state.region,
    data = ds,
    family = gaussian,
    corstr = "exchangeable"
  )

# tidier
broom::tidy(geefit, conf.int = TRUE)
#> # A tibble: 3 x 7
#>   term        estimate std.error statistic p.value conf.low conf.high
#>   <chr>          <dbl>     <dbl>     <dbl>   <dbl>    <dbl>     <dbl>
#> 1 (Intercept)  4406.      407.     117.      0      3608.     5205.  
#> 2 Frost           1.69      2.25     0.562   0.453    -2.72      6.10
#> 3 Murder        -22.7      31.4      0.522   0.470   -84.2      38.8

# glance
broom::glance(geefit)
#> Error in `$<-.data.frame`(`*tmp*`, "df.residual", value = 47L): replacement has 1 row, data has 0
```

<sup>Created on 2018-12-24 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>

jwilber:
Can I take this one over the weekend?

alexpghayes:
Go for it!

PR #580, opened 2019-01-12
#562 glance func for  geeglm

jwilber:
Hey all -
So my new `glance.geeglm` func will return the following (`geefit` defined as in #562):
```
> glance(geefit)
# A tibble: 1 x 5
  df.residual n.cluster max.cluster.size alpha   gamma
        <int>        <int>            <int> <dbl>   <dbl>
1          47           36                4 0.167 345510.
```

I built this from the output of`summary(geefit)`, which yields the following:
```
> summary(geefit)
Call:
geepack::geeglm(formula = Income ~ Frost + Murder, family = gaussian, 
    data = ds, id = state.region, corstr = "exchangeable")

 Coefficients:
            Estimate Std.err   Wald Pr(>|W|)    
(Intercept)  4406.19  407.45 116.94   <2e-16 ***
Frost           1.69    2.25   0.56     0.45    
Murder        -22.66   31.38   0.52     0.47    
---
Signif. codes:  0 ‘***’ 0.001 ‘**’ 0.01 ‘*’ 0.05 ‘.’ 0.1 ‘ ’ 1

Estimated Scale Parameters:
            Estimate Std.err
(Intercept)   345510   74689

Correlation: Structure = exchangeable  Link = identity 

Estimated Correlation Parameters:
      Estimate Std.err
alpha    0.167  0.0942
Number of clusters:   36   Maximum cluster size: 4 
```

**Currently, my glance test does not pass**. 
This is because the column names for my glance tibble (in particular,  `n.cluster`, `max.cluster.size`, `alpha`, & `gamma`)  are not in `column_glossary`. 
I couldn't find (from my knowledge, anyway) any equivalent column names. 
Should I rename these columns? Or should I submit a PR to edit `columns_glance.yaml` in `modeltests`?


alexpghayes:
Thanks, this looks good. You'll need to make a PR to `modeltests`. However, let's use `snake_case` for the column names:

- `n.cluster` -> `n_clusters`
- `max.cluster.size` -> `max_cluster_size`

I don't know much about `geeglm`, can someone chime in to let us know if we need to deal with different correlation structures differently? cc @IndrajeetPatel

jwilber:
Awesome - updated col names here and [here in `modeltests`.](https://github.com/alexpghayes/modeltests/pull/19)

IndrajeetPatil:
I am a bit confused by the column naming convention in `broom` now. As far as I have noticed, none of the column names from broom tidy or glance methods are in snake_case, right? 
(For example, 
`tidy`: p.value, std.error, conf.low, conf.high, etc.; 
`glance`:  df.residual, p.value, r.squared, adj.r.squared, etc.).

So why the sudden change in the naming convention? This becomes confusing for the user, or at least that's my opinion.

jwilber:
@IndrajeetPatil I have to agree here, and mentioned it in my PR for `modeltests`. 
I'll happily change it back if @alexpghayes agrees.

To be fair, the [tidyverse style guide]() isn't exactly clear on column-names:

On **object names**:
> Use underscores (_) (so called snake case) to separate words within a name.

On **dots**:

> Base R uses dots in function names (contrib.url()) and class names (data.frame), but it’s better to reserve dots exclusively for the S3 object system. In S3, methods are given the name function.class; if you also use . in function and class names, you end up with confusing methods like as.data.frame.data.frame().

That said, I think what's done in broom mirros the above: object names (e.g. `finish_glance`) use snake case, while columns (which are mostly just S3 methods) use dots.

I also think using snake case would be weird since in `augment` you'd require a mix of the two, e.g. `.std_resid` instead of `.std.resid`.
Just my two cents.

jwilber:
Any update on this?

alexpghayes:
I'm drowning in course work right now but am planning to spend some time on `broom` the weekend of March 9 - 10.

alexpghayes:
Okay, I agree with you both and think mixing underscores and periods is a bad idea. Let's revert to periods.

alexpghayes:
Sorry for all the delays -- merging now, will add dots back into names myself! Thank you for contributing!

jwilber:
@alexpghayes Thanks! Sorry, been busy w/ the end of the fiscal year. Thanks for doing what you do, I look forward to further PRs :)

Issue #487, opened 2018-08-15
tidy() gives a warning when used with anova()


AndrewKostandy:
After the latest broom update (0.5.0), the tidy function gives the warning: "Warning: Unknown or uninitialised column: 'term'." when used on the anova function. That was not the case before the update.

```r
m1 <- lm(mpg~hp, data=mtcars)
m2 <- lm(mpg~hp+wt, data=mtcars)
broom::tidy(anova(m1,m2))
#> Warning: Unknown or uninitialised column: 'term'.
#> # A tibble: 2 x 6
#>   res.df   rss    df sumsq statistic     p.value
#> *  <dbl> <dbl> <dbl> <dbl>     <dbl>       <dbl>
#> 1     30  448.    NA   NA       NA   NA         
#> 2     29  195.     1  253.      37.6  0.00000112
```

alexpghayes:
Hmmm, will look into this.

topepo:
I get a similar issue when using one of the updater functions that has a column for AIC (but maybe a different issue)

``` r
# from ?drop1
require(graphics); require(utils); require(broom); require(dplyr)
#> Loading required package: broom
#> Loading required package: dplyr
#> 
#> Attaching package: 'dplyr'
#> The following objects are masked from 'package:stats':
#> 
#>     filter, lag
#> The following objects are masked from 'package:base':
#> 
#>     intersect, setdiff, setequal, union
## following example(swiss)
lm1 <- lm(Fertility ~ ., data = swiss)
add1(lm1, ~ I(Education^2) + .^2) %>% tidy()
#> Warning in tidy.anova(.): The following column names in ANOVA output were
#> not recognized or transformed: AIC
#> # A tibble: 12 x 5
#>    term                            df  sumsq   rss   AIC
#>    <chr>                        <dbl>  <dbl> <dbl> <dbl>
#>  1 <none>                          NA  NA    2105.  191.
#>  2 I(Education^2)                   1  11.8  2093.  192.
#>  3 Agriculture:Examination          1  10.7  2094.  192.
#>  4 Agriculture:Education            1   1.83 2103.  193.
#>  5 Agriculture:Catholic             1  75.0  2030.  191.
#>  6 Agriculture:Infant.Mortality     1   4.44 2101.  193.
#>  7 Examination:Education            1  48.7  2056.  192.
#>  8 Examination:Catholic             1  40.8  2064.  192.
#>  9 Examination:Infant.Mortality     1  65.9  2039.  191.
[...]

Issue #558, opened 2018-12-17
allow `lmRob` and `glmRob` tidiers to take `conf.int` argument without error


IndrajeetPatil:
Currently, confidence intervals are not available for models from `robust` package, so I was expecting setting `conf.int = TRUE` would just return the tidy output without confidence intervals. But instead this actually produces an error.
``` r

# set up
set.seed(123)
library(robust)
#> Loading required package: fit.models
library(broom)
data(stack.dat)

# model
stack.rob <- lmRob(Loss ~ ., data = stack.dat)

# tidy output - with and without conf.int set to `TRUE`
broom::tidy(x = stack.rob)
#> # A tibble: 4 x 5
#>   term        estimate std.error statistic       p.value
#>   <chr>          <dbl>     <dbl>     <dbl>         <dbl>
#> 1 (Intercept) -37.7       5.00       -7.53 0.000000829  
#> 2 Air.Flow      0.798     0.0713     11.2  0.00000000291
#> 3 Water.Temp    0.577     0.175       3.29 0.00432      
#> 4 Acid.Conc.   -0.0671    0.0651     -1.03 0.318

broom::tidy(x = stack.rob, conf.int = TRUE)
#> Error in UseMethod("vcov"): no applicable method for 'vcov' applied to an object of class "lmRob"
traceback()
#> No traceback available
```

This is not the case for other tidiers which clearly don't output confidence intervals but nevertheless work if `conf.int = TRUE`. For example-
```r
# model for which confidence intervals are not going to be available
mtcars.chisq <- stats::chisq.test(x = mtcars$am, y = mtcars$cyl)
#> Warning in stats::chisq.test(x = mtcars$am, y = mtcars$cyl): Chi-squared
#> approximation may be incorrect

broom::tidy(x = mtcars.chisq, conf.int = TRUE)
#> # A tibble: 1 x 4
#>   statistic p.value parameter method                    
#>       <dbl>   <dbl>     <int> <chr>                     
#> 1      8.74  0.0126         2 Pearson's Chi-squared test
```

It will be nice to have this behavior also for tidiers for `robust` package models specifically, and other models more generally. This way, one doesn't need to keep changing the `broom::tidy()` function call based on the class of the object. 

<sup>Created on 2018-12-17 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>

<details>

<summary>Session info</summary>

``` r
devtools::session_info()
#> - Session info ----------------------------------------------------------
[...]

PR #583, opened 2019-01-19
fixing issue #487: ANOVA warning

brunaw:


brunaw:
There is actually two things being done here:

1. Fix #487: the ANOVA "term" warning 
2. Fix #558: now lmRob/glmRob accepts both `TRUE` and `FALSE` for `conf.int`, but will always return the results for `FALSE`

Issue #570, opened 2019-01-05
p-values for `lmodel2` are missing


IndrajeetPatil:
Summary for `lmodel2` objects include a column corresponding to _p_-values (`"P-perm (1-tailed)"`), but they are included in the tidier output.
``` r
# setup
set.seed(123)
library(lmodel2)
data(mod2ex2)

# model
Ex2.res <-
  lmodel2::lmodel2(
    formula = Prey ~ Predators,
    data = mod2ex2,
    range.y = "relative",
    range.x = "relative",
    nperm = 99
  )

# regression results
Ex2.res$regression.results
#>   Method Intercept    Slope Angle (degrees) P-perm (1-tailed)
#> 1    OLS  20.02675 2.631527        69.19283              0.01
#> 2     MA  13.05968 3.465907        73.90584              0.01
#> 3    SMA  16.45205 3.059635        71.90073                NA
#> 4    RMA  17.25651 2.963292        71.35239              0.01

# p-values are absent
broom::tidy(Ex2.res, conf.int = TRUE)
#> # A tibble: 8 x 5
#>   method term      estimate conf.low conf.high
#>   <fct>  <chr>        <dbl>    <dbl>     <dbl>
#> 1 MA     Intercept    13.1      1.35     19.8 
#> 2 MA     Slope         3.47     2.66      4.87
#> 3 OLS    Intercept    20.0     12.5      27.6 
#> 4 OLS    Slope         2.63     1.86      3.40
#> 5 RMA    Intercept    17.3      8.96     23.8 
#> 6 RMA    Slope         2.96     2.17      3.96
#> 7 SMA    Intercept    16.5      9.20     22.1 
#> 8 SMA    Slope         3.06     2.38      3.93
```

<sup>Created on 2019-01-05 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>

mattle24:
@IndrajeetPatil I can add the p-value to the `tidy.lmodel2` output. I don't know if it was a deliberate decision not to include it, so I'll have to wait on someone to weigh in on that.

alexpghayes:
Yeah, adding that in would be great. The goal of broom is not to be opinionated in what gets returned, just to clean model objects up.

mattle24:
Great, thanks Alex! 

New output looks like this:


``` r
# setup
set.seed(123)
library(lmodel2)
data(mod2ex2)

# model
Ex2.res <-
  lmodel2::lmodel2(
[...]

PR #592, opened 2019-01-22
Feature: Add p-values to tidy.lmodel2()

mattle24:
Addressing https://github.com/tidymodels/broom/issues/570


mattle24:
On the first build AppVeyor failed installing curl:

`Error: (converted from warning) cannot remove prior installation of package 'curl'`

alexpghayes:
Can you update the documentation to include that the p-value is one-tailed (?) and calculated via a permutation test (?), and possibly note why this is?

mattle24:
Sure!

mattle24:
I updated the documentation to note the one-tailed p-values and explained why they're calculated with permutation tests. After looking through the `lmodel2` documentation I'm still not sure why they're one-sided though.

R/lmodel2-tidiers.R
```diff
@@ -55,6 +61,7 @@ tidy.lmodel2 <- function(x, ...) {
 
   ret %>%
     inner_join(confints, by = c("method", "term")) %>% 
+    select(-p.value, dplyr::everything()) %>%
```
alexpghayes: Why do you need the `select()` line here? What does it do?

R/stats-lm-tidiers.R
```diff
@@ -279,5 +279,52 @@ process_lm <- function(ret, x, conf.int = FALSE, conf.level = .95,
...
+  process_lm(ret, x, conf.int = conf.int, conf.level = conf.level, 
+             exponentiate = exponentiate)
+}
+
+#' TODO: document
+#' @rdname tidy.lm.beta
+#' @export
+tidy.summary.lm.beta <- function (x, ...) {
+  co <- stats::coef(x)
+  nn <- c("estimate", 'standardized.estimate', "std.error", "statistic", "p.value")
```
alexpghayes: Let's use `std.estimate` instead of `standardized.estimate` for the column name.

R/stats-lm-tidiers.R
```diff
@@ -279,5 +279,52 @@ process_lm <- function(ret, x, conf.int = FALSE, conf.level = .95,
...
+#' CarRegS = lm.beta::lm.beta(CarReg)
+#' 
+#' # Table format
+#' tidy(CarRegS)
+#'
+#' @export
+#' @family lm tidiers
+tidy.lm.beta <- function (x, conf.int = FALSE, conf.level = 0.95, exponentiate = FALSE, 
+                          quick = FALSE, ...) {
+  if (quick) {
```
alexpghayes: This looks like a copy-paste of `tidy.lm()`, which is a very reasonable place to start. However, `tidy.lm()` is a mess and is way more complicated than it needs to be. Would you mind attempting something more self-contained in the vein of `tidy.nls()` for example (https://github.com/tidymodels/broom/blob/master/R/stats-nls-tidiers.R#L32)?

alexpghayes:
The `lmodel2` stuff looks great. Left some comments on the `lm.beta` tidiers!

R/lmodel2-tidiers.R
mattle24: This changes the order of the columns so p.value is the rightmost column.

mattle24:
Sorry about all of this. I think I had an old and bad commit leftover on my forked repo's master. I pulled instead of reseting to the tidymodels/master, and that caused a bunch of problem. To fix this, I did some hard resets, so things may look a little funky now. I took out the lm.beta things though, all that should be here is the lmodel2 stuff.

alexpghayes:
Looks good, thanks!

Issue #589, opened 2019-01-21
Write tidy.lmRob() and tidy.glmRob() methods instead of using tidy.lm() on (g)lmRob objects


IndrajeetPatil:
-  Although PR to fix #558 solved the `conf.int` issue for `lmRob`, the same issue still holds for `glmRob`- 
``` r
# setup
set.seed(123)
library(robust)
#> Loading required package: fit.models
data(breslow.dat)

# tidy output
broom::tidy(
  x = robust::glmRob(
    sumY ~ Age10 + Base4 * Trt,
    family = poisson(),
    data = breslow.dat,
    method = "cubif"
  ),
  conf.int = TRUE
)
#> [1] "Variance number 1 smaller than 1.e-10 (set to 1.e-10)"
#> [1] "Variance number 2 smaller than 1.e-10 (set to 1.e-10)"
#> [1] "Variance number 3 smaller than 1.e-10 (set to 1.e-10)"
#> [1] "Variance number 4 smaller than 1.e-10 (set to 1.e-10)"
#> [1] "Variance number 5 smaller than 1.e-10 (set to 1.e-10)"
#> Error in UseMethod("vcov"): no applicable method for 'vcov' applied to an object of class "glmRob"
```
- Although these two tidiers take `quick` argument, if you specify it, the function produces error-
```r
# set up
set.seed(123)
library(robust)
#> Loading required package: fit.models
data(stack.dat)

# model
stack.rob <- robust::lmRob(formula = Loss ~ ., data = stack.dat)

# broom outputs
broom::tidy(x = stack.rob,
            conf.int = TRUE,
            conf.level = 0.95,
            quick = FALSE)
#> Warning in if (quick) {: the condition has length > 1 and only the first
#> element will be used
#> Error in if (quick) {: argument is not interpretable as logical
```
<sup>Created on 2019-01-21 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>

<details>

<summary>Session info</summary>

``` r
devtools::session_info()
#> - Session info ----------------------------------------------------------
#>  setting  value                                             
#>  version  R Under development (unstable) (2018-11-30 r75724)
#>  os       Windows 10 x64                                    
#>  system   x86_64, mingw32                                   
#>  ui       RTerm                                             
#>  language (EN)                                              
#>  collate  English_United States.1252                        
#>  ctype    English_United States.1252                        
#>  tz       Asia/Calcutta                                     
#>  date     2019-01-21                                        
#> 
[...]

PR #593, opened 2019-01-24
partial fix for #589

IndrajeetPatil:
`glmRob` tidier defaults to `conf.int = FALSE`
This PR doesn't address the issue with `quick` argument (in #589)


alexpghayes:
As a temporary patch this is fine, but really we need new `orcutt` and `*Rob` tidiers. There are now issues out for both of those.

Issue #82, opened 2015-10-26
Including number of observations used to build the model


mschubert:
I think it would be useful if some version of the `data.frame` that represents the result also include a column with the number of observations that were used to build the model.

An easy way to access would be using e.g. `nobs()` for `stats::lm()`, but I'm sure other models have similar reporting of this.

It would be even more useful if it could include the actual number of observations for `logicalTRUE` or `factorLEVEL`.

grasshoppermouse:
I also think it would be useful if glance output included nobs. It looks like glance.stanreg already does:

https://github.com/dgrtwo/broom/blob/master/R/rstanarm_tidiers.R

hughjonesd:
Just adding my vote to this feature. I think in most fields, it is standard to report the N as one of the summary statistics.

randomgambit:
hello there! is there a fix for that very important feature?
thanks!!

alexpghayes:
I'd be willing to this in a tryCatch to `finish_glance` to pick up an `n` column for relevant models. The question then is how many models actually implement `nobs()` methods.

I'm hesitant to report counts for each factor level in `glance()` because this is moving more into data summarizing than properties of a model. If a `tidy()` method doesn't already inherit this information from `summary()` or whatnot I don't think it's worth the effort to try implement this consistently across tidiers. Also, `skimr::skim()` is fantastic for this sort of thing.

vincentarelbundock:
@alexpghayes One possible design that I would be willing to implement:

For each element of the list above which are *not* compatible with `nobs`, extract the value explicitly in the model-specific glance function.

Modify the `finish_glance` function. If `'n' %in% names(ret)`, then `tryCatch(stats::nobs)`

This is a bit more work (which I am willing to do), but it's explicit, and would avoid unexpected side effects from having a bunch of ifelse statements.

vincentarelbundock:
See https://github.com/tidymodels/broom/pull/594

gavinsimpson:
Sounds simpler to just implement `nobs()` methods for these internally, where they don't exist. Ideally these would be offered upstream to the respective package maintainers, but there's nothing stopping these being in broom if they are not wanted or maintainers are unresponsive. Using `nobs()` makes it simpler/safer to implement return of number of observations in `glance`.

vincentarelbundock:
[...]

PR #594, opened 2019-01-24
WIP: Add number of observations to `glance` output.

vincentarelbundock:
Where `n` is not specified explicitly in the model-specific `glance` function, `finish_glance` tries to extract the number of observations using `stats::nobs`.

This addresses: https://github.com/tidymodels/broom/issues/82

### DIFF
--- a/R/nnet-tidiers.R
+++ b/R/nnet-tidiers.R
@@ -106,7 +106,8 @@ glance.multinom <- function(x, ...) {
     tibble(
       edf = edf,
       deviance = deviance,
-      AIC = AIC
+      AIC = AIC,
+      n = nrow(x$residuals)
     )
   )
 }
--- a/R/plm-tidiers.R
+++ b/R/plm-tidiers.R
@@ -81,7 +81,8 @@ glance.plm <- function(x, ...) {
     r.squared = r.squared[1],
     adj.r.squared = r.squared[2],
     statistic = fstatistic$statistic,
-    p.value = fstatistic$p.value
+    p.value = fstatistic$p.value,
+    n = length(s$residuals)
   ))
   finish_glance(ret, x)
 }
--- a/R/robust-glmrob-tidiers.R
+++ b/R/robust-glmrob-tidiers.R
@@ -63,7 +63,8 @@ augment.glmRob <- function(x, ...) {
 glance.glmRob <- function(x, ...) {
   ret <- tibble(
     deviance = x$deviance,
-    null.deviance = x$null.deviance
+    null.deviance = x$null.deviance,
+    n = length(x$residuals)
   )
   finish_glance(ret, x)
 }
--- a/R/robust-lmrob-tidiers.R
+++ b/R/robust-lmrob-tidiers.R
@@ -92,6 +92,7 @@ glance.lmRob <- function(x, ...) {
     r.squared = x$r.squared,
     deviance = x$dev,
     sigma = s$sigma,
-    df.residual = x$df.residual
+    df.residual = x$df.residual,
+    n = length(x$residuals)
   )
 }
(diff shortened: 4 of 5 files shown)


alexpghayes:
I like the idea of this.

My primary concern is that a bunch of `glance` methods just gained an undocumented column in their return. If people are indexing columns based on position, this could introduce breaking changes. I'd really like to get rid of `finish_glance()` because it just tries a bunch of stuff and returns what worked, but it's never really clear what will work for a particular model. The downside is that this means every model object needs its own glance method, which is no small task.

cc @dgrtwo any thoughts?

vincentarelbundock:
Thanks for looking at this and commenting so quickly!

Just to push back a bit, I'm not sure the "undocumented column" issue should be enough to reject this. Columns are already very different from model to model, so it's not like there is *one* standard set of columns that users can expect from model to model. There is no "documented" set of columns *per se*.

Of course, you are right that this could be a breaking change for some people who manipulate based on column positions. However, taking this argument seriously would prevent `broom` from *ever* changing the statistics that are presented in any model or the order in which those statistics are presented (shouldn't they be alphabetical?). In my own practice, I have found that *N* is probably the *single most* common/important model feature that is reported in scientific communication. If we are unwilling to modify `glance` output for that, then it's game over.

Moreover, I totally get the "non-breaking" changes argument. But playing with objects produced by an external dependency like broom by column number would be *terrible* coding practice. Obviously, it's great that you care enough to protect users against their own worst instincts, but there's a limit to the extent of awfulness you might want to guard against. “Bad code” = OK; “Terrible code” = not OK. Otherwise, it’s _status quo_ forever.

I understand why you don’t feel comfortable with `finish_glance`: It’s not as explicit as perhaps would be ideal. But this just seems like a harsh trade-off against the other sin: code duplication. Given the growing scope of `broom`, maintainability is likely to become one of the main challenges of this community. In that context, limiting code duplication seems like a good call.

alexpghayes:
Those are good points. To be clear, I *do* want to add `nobs` to `glance()`, I just want to make sure we do it in a way that doesn't causes surprises / breakages / is hard to maintain.

> Just to push back a bit, I'm not sure the "undocumented column" issue should be enough to reject this. Columns are already very different from model to model, so it's not like there is one standard set of columns that users can expect from model to model. There is no "documented" set of columns per se.

Right, but if you enter `?glance.lm`, you'll see which columns the `glance.lm()` returns. Updating `finish_glance()` will necessitate updating the `@return` documentation for many models. Which we totally can do, it's just a bunch of work.

> I understand why you don’t feel comfortable with finish_glance: It’s not as explicit as perhaps would be ideal. But this just seems like a harsh trade-off against the other sin: code duplication. Given the growing scope of broom, maintainability is likely to become one of the main challenges of this community. In that context, limiting code duplication seems like a good call.

Oddly enough, I've found that having individualized methods is more maintainable, mostly because it isolates different parts of the code base and prevents unexpected ramifications. In the end, I'd really like to do with away with `finish_glance()` entirely, but my guess is that this would take about a full week of work to accomplish. In the meantime, I like approach you're taking. Let's just make sure that we:

- appropriately test for the `nobs` column presence so we can catch regression errors if `finish_glance()` does get split up down the line
- document columns like I mentioned above

The gist you linked to in the issue thread is super useful! Thanks for taking this on!

alexpghayes:
That's fantastic! Thank you!

> Host nobs method inside broom for the objects types where none is defined. (Eventually, we can offload those upstream if there are takers.)

Writing new `nobs` methods is probably overkill. I'd just add model-specific code in the `glance.*` methods.

> Add an n column explicitly from inside the model-specific glance.

> Document n using roxygen2 for the models where it is added.

Sounds good.

> Move the LogLik, AIC, and BIC calls to model-specific glances. Get rid of finish_glance once 1-4 are completed for all models.

If you want to do this, more power to you, but I'm also happy to merge without this change.

vincentarelbundock:
Closing this PR. Please review this one instead: https://github.com/tidymodels/broom/pull/597

PR #597, opened 2019-01-28
remove finish_glance and add nobs

vincentarelbundock:
Following up on https://github.com/tidymodels/broom/pull/594, I wanted to see how much work it would be to replace `finish_glance` with individual calls to `stats::BIC` et al. inside every glance method (with `@evalRd return_glance`). Turned out not to be that bad (well, it was pretty bad, but I'm quite far along now). I also improved the docs in some places along the way.

Testing locally is difficult for some reason, so I'm opening this WIP PR just to see how far I am to a clean change.


vincentarelbundock:
This PR does the following:

1. Removes `finish_glance` completely
2. Moves the `AIC, BIC, logLik, deviance, df.residual` calls to model-specific glance methods. 
3. Add `nobs` using `stats::nobs` to the vast majority of glance methods (all the ones I could figure out). Wherever possible, `nobs` is added last to preserve the column order of the other glance entries.
4. Adds a `R/nobs.R` file which includes `nobs` methods for models that do not export one.
5. Minor updates to the tests. Mostly, this involved changing the hard-coded number of columns in glance output.

Travis gives us the OK.

There are two doc-related warnings which might have been introduced by this PR. I can't quite figure out how to make them disappear. Any tips would be much appreciated.

```
* checking Rd files ... WARNING
prepare_Rd: lme4_tidiers.Rd: Only one \value section is allowed: the first will be used
prepare_Rd: ordinal_tidiers.Rd: Only one \value section is allowed: the first will be used
```

@alexpghayes Please let me know if there's anything you'd like me to improve.

alexpghayes:
Will give this a thorough look in a day or two!

alexpghayes:
Okay, so I'm definitely on grad student time and I apologize for that! 

This is fantastic! Will you:

- [ ] Add yourself as a contributor in `DESCRIPTION`
- [ ] Add a note that many `glance` methods now return the number of observations in an `nobs` column, which is typically the rightmost column

vincentarelbundock:
cool cool. I did those two things and fixed some minor conflicts against master. Also removed lme4 and nlme4 which appear to have been deprecated.

Travis breaks, but two of their machines break and `lavaan`, which I have not touched, and one says `robust` package missing.

Here's what the tests look like on my machine: https://gist.github.com/vincentarelbundock/dee3cb1dbb3634c442f595f3ddc39310

alexpghayes:
Thank you so much! Will clean up breakages myself!

topepo:
Were reverse dependencies checked for this change? There are some packages that will break such as [`sweep`](https://github.com/business-science/sweep/blob/5a3fc5d0bd07d18ba16bc0cfea60bb9dbad310c2/R/tidiers_arima.R#L94). @mdancho84

alexpghayes:
That was expected. This is a very good change.

simonpcouch:
There are a couple reverse dependencies broken, yes (#862). Filing issues/PRs right now!

PR #610, opened 2019-02-26
Add a tidy.systemfit() method

jaspercooper:
This PR adds tidy methods that should produce tibbles for any model of class [`systemfit`](https://cran.r-project.org/web/packages/systemfit/systemfit.pdf). I am not a developer of that package.

- [x] tests added and coverage for systemfit is at 100%
- [x] `devtools::check()` produces no warnings, errors, or notes introduced by this PR (`1 error  | 2 warnings | 2 notes`)
- [x] tidy method documented according to guidelines 
- [x] added myself as a contributor
- [x] added my wonderful teammate @medinali as a contributor where she had originally added herself for her `tidy.mediate` method

An example that demonstrates the new behavior implemented in my PR:

```
rm(list = ls())
library(broom)
library(systemfit)
data( "Kmenta" )
eqDemand <- consump ~ price + income
eqSupply <- consump ~ price + farmPrice + trend
system <- list( demand = eqDemand, supply = eqSupply )

## iterated SUR estimation
fitsur <- systemfit( system, "SUR", data = Kmenta, maxit = 100 )
tidy( fitsur )

## 2SLS estimation
inst <- ~ income + farmPrice + trend
fit2sls <- systemfit( system, "2SLS", inst = inst, data = Kmenta )
tidy( fit2sls )
```

Let's see what surprises our friends at `appveyor` and Travis have in store for us! 

Thanks to the `broom` team for a wonderful package.


R/systemfit-tidiers.R
```diff
@@ -0,0 +1,52 @@
+#' @templateVar class systemfit
+#' @template title_desc_tidy
+#' 
+#' @param x A `systemfit` object produced by a call to [systemfit::systemfit()].
+#' @param conf.int Logical indicating whether or not to include a confidence interval in the tidied output. Defaults to `FALSE`.
```
alexpghayes: Can you add an additional argument `conf.level` defaulting to 0.95 to allow the use to set the confidence level? Then you can simultaneously document both arguments using the `@template param_confint` to achieve consistent documentation!

R/systemfit-tidiers.R
```diff
@@ -0,0 +1,52 @@
...
+  sf_coefs <- sf_coefs[,- which(names(sf_coefs) == "t value")]
+  sf_coefs <- cbind(term = rownames(sf_coefs), sf_coefs)
+  if(conf.int){
+    nn <- c(nn, "conf.low", "conf.high")
+    sf_cis <- confint(x)
+    sf_cis <- matrix(sf_cis,ncol = 2)
+    sf_coefs <- cbind(sf_coefs, sf_cis)
+  }
+  names(sf_coefs) <- nn
+  rownames(sf_coefs) <- NULL
```
alexpghayes: Setting `rownames` shouldn't be necessary. Also, if you define `nn` here it'll be a bit easy to read this code.

R/systemfit-tidiers.R
```diff
@@ -0,0 +1,52 @@
...
+#' @export
+#' @seealso [tidy()], [systemfit::systemfit()]
+#' @family systemfit tidiers
+#' @aliases systemfit_tidiers
+tidy.systemfit <- function(x, conf.int = TRUE, ...) {
+  nn <- c("term", "estimate", "std.error", "p.value")
+  
+  sf_summary <- summary(x)
+  sf_coefs <- as.data.frame(sf_summary$coefficients)
+  sf_coefs <- sf_coefs[,- which(names(sf_coefs) == "t value")]
```
alexpghayes: Is there a reason to strip out the T statistic? If not, let's keep it and call that column `statistic`.

R/systemfit-tidiers.R
```diff
@@ -0,0 +1,52 @@
...
+  sf_coefs <- cbind(term = rownames(sf_coefs), sf_coefs)
+  if(conf.int){
+    nn <- c(nn, "conf.low", "conf.high")
+    sf_cis <- confint(x)
+    sf_cis <- matrix(sf_cis,ncol = 2)
```
alexpghayes: Just curious -- what happens without this line? Is it necessary to coerce to matrix since we're going to end up as a `tibble` anyway?

alexpghayes:
Wow this is fantastic! Thanks for the contribution!

I left a couple comments on the code for you to take a look at when you have a moment!

R/systemfit-tidiers.R
jaspercooper: This was actually a deliberate omission. Totally happy to add this, but didn't originally because `conf.level` doesn't change the CIs in the method that they've created for it in the original systemfit package, so it seemed a bit disingenuous to tell users they could input this. LMK your preference

R/systemfit-tidiers.R
jaspercooper: Done in 74dc5946df85ba944cd202ce7a7210ae5cf75cac

R/systemfit-tidiers.R
jaspercooper: Done in 31b55e4b2bc776f2f36b6fa3d2c3959c0c624824

R/systemfit-tidiers.R
jaspercooper: Without it you get
> Error in as.data.frame.default(x[[i]], optional = TRUE) : 
>  cannot coerce class ‘"confint.systemfit"’ to a data.frame
It's a really weird object and very hard to wrangle:
```r
> str(sf_cis)
 'confint.systemfit' num [1:4, 1:2] -0.0339 -0.2893 -0.2527 -0.186 0.4179 ...
 - attr(*, "dimnames")=List of 2
  ..$ : chr [1:4] "eq1_(Intercept)" "eq1_Z" "eq2_(Intercept)" "eq2_X"
  ..$ : chr [1:2] "2.5 %" "97.5 %"
```
LMK if you want to take a different approach

jaspercooper:
Hi @alexpghayes, thanks for the review! 

I've implemented many of your suggestions and left thought on a couple others. Have also updated the branch to not have conflicts with master.

R/systemfit-tidiers.R
alexpghayes: Would you add this? Unless it totally isn't possible, in which case you should set `strict = FALSE` in the `check_arguments()`  test and document that this is the case.

R/systemfit-tidiers.R
alexpghayes: No problem, let's stick with this.

alexpghayes:
Sorry for the delay, just took another look!

R/systemfit-tidiers.R
jaspercooper: Added.

jaspercooper:
OK good to go when you are, @alexpghayes

alexpghayes:
Thanks for making this happen!

PR #651, opened 2019-03-10
use travis to build website

IndrajeetPatil:
Next, make sure that a `gh-pages` branch exists. The simplest way to do so is to run the following git commands locally:

```bash
git checkout --orphan gh-pages
git rm -rf .
git commit --allow-empty -m 'Initial gh-pages commit'
git push origin gh-pages
git checkout master
```

We recommend doing this outside of RStudio (with the project closed) as from RStudio's perspective you end up deleting all the files and then re-creating them.


alexpghayes:
Hey @jimhester can we get a sanity check that this is the right way to set up automatic pkgdown deploys?

jimhester:
I think that seems right, @alexpghayes you will need to run `travis::use_travis_deploy()` from the [travis package](https://github.com/ropenscilabs/travis) to setup the necessary deploy key so you can push to the gh-pages branch to deploy the site.

alexpghayes:
@IndrajeetPatil does the build need to pass for the site to deploy?

IndrajeetPatil:
No, it should deploy even if the build doesn't succeed. 

Only the parts of the `yaml` which are in the `after_success` entry will work after the builds succeed.
https://github.com/tidymodels/broom/blob/ae40a4b92530a1205a5bd56975acecfbe213058a/.travis.yml#L71-L74

alexpghayes:
Any idea why Travis won't build https://travis-ci.org/tidymodels/broom/builds/517607964#L3476? Is this related to what you added or something else?

PR #653, opened 2019-03-10
Add conf.int functionality for tidy.gamlss()

IndrajeetPatil:
- This catches up with tidier for gamlss in broom.mixed
- Added few more tests for this one
- Few additional changes to avoid warnings


R/gamlss-tidiers.R
```diff
@@ -1,39 +1,53 @@
...
-#'   parameter = "Type of coefficient being estimated: `mu`, `sigma`,
-#'     `nu`, or `tau`.",
-#'   "term",
-#'   "estimate",
-#'   "std.error",
-#'   "statistic",
-#'   "p.value"
-#' )
-#' 
+#' Tidying methods for gamlss objects
```
alexpghayes: Much of this documentation change is a regression to the old style.

R/gamlss-tidiers.R
```diff
@@ -1,39 +1,53 @@
...
-#'   y ~ pb(x),
-#'   sigma.fo = ~ pb(x),
-#'   family = BCT,
-#'   data = abdom,
-#'   method = mixed(1, 20)
-#' )
-#'
-#' tidy(g)
+#' \dontrun{
+#' if (requireNamespace("gamlss", quietly = TRUE) &&
```
alexpghayes: The examples as well -- let's keep what was in `broom`. The only new thing here is checking if `gamlss` is loaded. Technically you should do this, but it makes the examples harder to read and is never the case in practice, so the standard in `broom` is not to do this.

R/gamlss-tidiers.R
```diff
@@ -1,39 +1,53 @@
...
+
+tidy.gamlss <- function(x, quick = FALSE, conf.int = FALSE, conf.level = 0.95, ...) {
   if (quick) {
     co <- stats::coef(x)
-    return(tibble(term = names(co), estimate = unname(co)))
+    return(dplyr::tibble(term = names(co), estimate = unname(co)))
+  }
+
+  # need gamlss for summary to work
+  if (!requireNamespace("gamlss", quietly = TRUE)) {
```
alexpghayes: I guess we can keep this but we don't do this anywhere else in `broom`. At some point I'll write a helper function that makes this a nice single function call.

R/gamlss-tidiers.R
```diff
@@ -43,8 +57,25 @@ tidy.gamlss <- function(x, quick = FALSE, ...) {
   nn <- c("estimate", "std.error", "statistic", "p.value")
   ret <- fix_data_frame(s, nn)
 
-  # add parameter types. This assumes each coefficient table starts
-  # with "(Intercept)": unclear if this is guaranteed
-  parameters <- x$parameters[cumsum(ret$term == "(Intercept)")]
+  if (conf.int) {
```
alexpghayes: The confidence interval tidying appears to be the meat of the new code.

R/gamlss-tidiers.R
```diff
@@ -43,8 +57,25 @@ tidy.gamlss <- function(x, quick = FALSE, ...) {
...
+    x$parameters,
+    vapply(coefs, length, 1L)
+  )
   bind_cols(parameter = parameters, ret)
 }
+
+#' @export
+glance.gamlss <- function(x, ...) {
+  ret <- tibble(df = x$df.fit)
+  finish_glance(ret, x)
```
alexpghayes: `finish_glance()` has also been removed from `broom` so you'll need to recreate whatever specific functionality from that that's getting used here.

man-roxygen/boilerplate.R
```diff
@@ -0,0 +1,2 @@
+#' @return All tidying methods return a data.frame without rownames, whose
```
alexpghayes: This boilerplate is also from the old documentation strategy, and I don't want to revert.

tests/testthat/test-gamlss.R
```diff
@@ -13,14 +13,17 @@ library(gamlss.dist)
 
 data(abdom, package = "gamlss.data")
 
-fit <- gamlss(
+# all 50 warnings are the same
```
alexpghayes: Good to know!

alexpghayes:
Are you not interested any more or have I just been too slow to respond to `broom` stuff recently?

IndrajeetPatil:
Hey Alex,

I closed this because I will not be able to work on this. Additionally, given the diverging core structure and philosophies between the two packages, I think this tidier needs to. E written from scratch, or at least should not be copy pasted. 

Maybe someone who actually works with these models and has more experience than me can handle this.

Issue #603, opened 2019-02-11
augment method for `nls` not working with `se_fit = TRUE`


IndrajeetPatil:
``` r
# setup
set.seed(123)
library(graphics)
DNase1 <- subset(DNase, Run == 1)

# model
fm1DNase1 <- nls(density ~ SSlogis(log(conc), Asym, xmid, scal), DNase1)

# augmented dataframe
broom::augment(x = fm1DNase1, se_fit = FALSE)
#> # A tibble: 16 x 3
#>    density    conc .fitted
#>      <dbl>   <dbl>   <dbl>
#>  1   0.017  0.0488  0.0307
#>  2   0.018  0.0488  0.0307
#>  3   0.121  0.195   0.112 
#>  4   0.124  0.195   0.112 
#>  5   0.206  0.391   0.209 
#>  6   0.215  0.391   0.209 
#>  7   0.377  0.781   0.374 
#>  8   0.374  0.781   0.374 
#>  9   0.614  1.56    0.633 
#> 10   0.609  1.56    0.633 
#> 11   1.02   3.12    0.981 
#> 12   1.00   3.12    0.981 
#> 13   1.33   6.25    1.37  
#> 14   1.36   6.25    1.37  
#> 15   1.73  12.5     1.71  
#> 16   1.71  12.5     1.71

broom::augment(x = fm1DNase1, se_fit = TRUE)
#> Error: $ operator is invalid for atomic vectors

> traceback()
3: augment_newdata(x, data, newdata, se_fit)
2: augment.nls(fm1DNase1, se_fit = TRUE)
1: augment(fm1DNase1, se_fit = TRUE)
```

<sup>Created on 2019-02-10 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>

<details>

<summary>Session info</summary>

``` r
devtools::session_info()
#> - Session info ----------------------------------------------------------
#>  setting  value                                             
#>  version  R Under development (unstable) (2018-11-30 r75724)
#>  os       Windows 10 x64                                    
#>  system   x86_64, mingw32                                   
#>  ui       RTerm                                             
#>  language (EN)                                              
#>  collate  English_United States.1252                        
#>  ctype    English_United States.1252                        
#>  tz       America/New_York                                  
#>  date     2019-02-10                                        
#> 
#> - Packages --------------------------------------------------------------
#>  package     * version    date       lib source                           
[...]

Issue #605, opened 2019-02-11
augment method for `nlrq` not working with `se_fit = TRUE` 


IndrajeetPatil:
``` r
# data
set.seed(123)
library(quantreg)
#> Loading required package: SparseM
#> 
#> Attaching package: 'SparseM'
#> The following object is masked from 'package:base':
#> 
#>     backsolve
Dat <- NULL
Dat$x <- rep(1:25, 20)
Dat$y <- SSlogis(Dat$x, 10, 12, 2) * rnorm(500, 1, 0.1)

# model
Dat.nlrq <-
  nlrq(
    y ~ SSlogis(x, Asym, mid, scal),
    data = Dat,
    tau = 0.5,
    trace = TRUE
  )
#> 111.7875 :   9.972779 11.925654  1.953960 
#> final  value 111.478251 
#> converged
#> lambda = 1 
#> 111.4783 :   9.902380 11.887034  1.946602 
#> final  value 111.253442 
#> converged
#> lambda = 1 
#> 111.2534 :   9.850718 11.856067  1.951539 
#> final  value 111.204692 
#> converged
#> lambda = 0.8714618 
#> 111.2047 :   9.810413 11.838020  1.953014 
#> final  value 111.197998 
#> converged
#> lambda = 1 
#> 111.198 :   9.816534 11.841982  1.958020 
#> final  value 111.197732 
#> converged
#> lambda = 1 
#> 111.1977 :   9.817044 11.839574  1.957293 
#> final  value 111.197675 
#> converged
#> lambda = 0.8741032 
#> 111.1977 :   9.817114 11.838862  1.957161 
#> final  value 111.197698 
#> converged
#> lambda = 0.9999188 
#> 111.1977 :   9.817114 11.838893  1.956801

# augmented dataframe
broom::augment(x = Dat.nlrq)
#> # A tibble: 500 x 3
#>        x      y .fitted
#>    <int>  <dbl>   <dbl>
#>  1     1 0.0384  0.0384
#>  2     2 0.0654  0.0639
#>  3     3 0.127   0.106 
#>  4     4 0.181   0.176 
#>  5     5 0.297   0.289 
#>  6     6 0.556   0.473 
#>  7     7 0.794   0.764 
#>  8     8 1.04    1.21  
#>  9     9 1.70    1.86  
#> 10    10 2.57    2.76  
#> # ... with 490 more rows

# add `.se.fit` column 
broom::augment(x = Dat.nlrq, se_fit = TRUE)
[...]

PR #654, opened 2019-03-10
[BUG] nls augment se_fit param

cwang23:
Resolving [this](https://github.com/tidymodels/broom/issues/603) issue.


alexpghayes:
I AM SORRY THIS TOOK SO LONG THANK YOU

IndrajeetPatil:
@cwang23 Do you also want to take a stab at https://github.com/tidymodels/broom/issues/605? It is very much the same issue and the solution is going to be pretty similar as well.

Issue #581, opened 2019-01-12
`boot` tidier doesn't work for some `conf.method` types


IndrajeetPatil:
Tidier for `boot` object works inconsistently. 

Here is a custom function to get confidence interval for Spearman's rho using `boot` package.
``` r
# setup
set.seed(123)
library(tidyverse)

# custom function
cor_ci <- function(data,
                   x,
                   y,
                   method = "spearman",
                   exact = FALSE,
                   continuity = TRUE,
                   alternative = "two.sided",
                   nboot = 100,
                   conf.level = 0.95,
                   conf.type = "norm",
                   ...) {
  # creating a dataframe from entered data
  data <- dplyr::select(
    .data = data,
    x = !!rlang::enquo(x),
    y = !!rlang::enquo(y)
  ) %>%
    dplyr::filter(.data = ., !is.na(x), !is.na(y)) %>%
    tibble::as_tibble(x = .)

  # function to obtain 95% CI for xi
  corci <- function(formula,
                      x,
                      y,
                      method = method,
                      exact = exact,
                      continuity = continuity,
                      alternative = alternative,
                      indices) {
    # allows boot to select sample
    d <- data[indices, ]
    # allows boot to select sample
    boot_df <-
      broom::tidy(
        stats::cor.test(
          formula = stats::as.formula(~ x + y),
          data = d,
          method = method,
          exact = exact,
          continuity = continuity,
          alternative = alternative,
          na.action = na.omit
        )
      )

    # return the value of interest: effect size
    return(boot_df$estimate)
  }

  # save the bootstrapped results to an object
  bootobj <- boot::boot(
    data = data,
    statistic = corci,
    R = nboot,
    x = x,
    y = y,
    method = method,
    exact = exact,
    continuity = continuity,
    alternative = alternative,
    parallel = "multicore",
    ...
  )

  # tidy dataframe with confidence intervals
  df <- broom::tidy(
    bootobj,
    conf.int = TRUE,
    conf.level = conf.level,
    conf.method = conf.type
  )

  # get 95% CI from the bootstrapped object
  bootci <- boot::boot.ci(
    boot.out = bootobj,
    conf = conf.level,
    type = conf.type
  )

  # return both of these objects
  list(df, bootci)
}
```

- **basic**: works

```r
cor_ci(mtcars, wt, mpg, conf.type = "basic")

#> [[1]]
#> # A tibble: 1 x 5
#>   statistic    bias std.error conf.low conf.high
[...]

PR #655, opened 2019-03-10
Fixed bug with norm bootstrap CIs and added error message for stud bootstrap CIs

krauskae:
Issue #581


alexpghayes:
Looks great. Can you:

- [ ] Add a test to make sure we're solved the problem
- [ ] Update `NEWS.md` in the Bug Fix section to tell other people we fixed this
- [ ] Add yourself as a contributor in `DESCRIPTION`
- [ ] Use `arg_match` on `conf.method`

PR #658, opened 2019-03-10
Separate tidy.mlm() and tidy.lm() methods. Fixes #510.

### LINKED ISSUE #510: tidy on mlm with conf.int = TRUE returns duplicated rows
`tidy(mlm_object, conf.int = TRUE)` returns duplicated values for the confidence intervals. I looked at the code and could not figure out why.

---

``` r
library(broom)
packageVersion("broom")
#> [1] '0.5.0.9001'

fit_mlm <- lm(cbind(mpg, disp) ~ wt, mtcars)
tidy(fit_mlm, conf.int = TRUE)
#> # A tibble: 4 x 8
#>   response term   estimate std.error statistic  p.value conf.low conf.high
#>   <chr>    <chr>     <dbl>     <dbl>     <dbl>    <dbl>    <dbl>     <dbl>
#> 1 mpg      (Inte…    37.3      1.88      19.9  8.24e-19    33.5      41.1 
#> 2 mpg      wt        -5.34     0.559     -9.56 1.29e-10    -6.49     -4.20
#> 3 disp     (Inte…  -131.      35.7       -3.67 9.33e- 4    33.5      41.1 
#> 4 disp     wt       112.      10.6       10.6  1.22e-11    -6.49     -4.20
```

<sup>Created on 2018-10-11 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>

IndrajeetPatil:


R/stats-lm-tidiers.R
```diff
@@ -57,29 +57,31 @@
...
-    if (inherits(x,'mlm')) {
-      ret <- data.frame(response = rep(colnames(co), each = nrow(co)),
-                        term = rep(rownames(co), times = ncol(co)),
-                        estimate = as.numeric(co), stringsAsFactors = FALSE)
-    } else {
-      ret <- data.frame(term = names(co), estimate = unname(co),
-                        stringsAsFactors = FALSE)
-    }
+
+    ret <- data.frame(
```
alexpghayes: Let's use `tibble` here instead of `data.frame(..., stringsAsFactors = FALSE)`

R/stats-lm-tidiers.R
```diff
@@ -219,33 +218,11 @@ glance.lm <- function(x, ...) {
   )
 }
 
-# compute confidence intervals for mlm object. 
-confint.mlm <- function (object, level = 0.95, ...) {
```
alexpghayes: Just a note: `confint.mlm()` is new in R 3.5. So removing this means we'll get some error on older versions of R that we should catch / do something informative with. i.e. I think deleting this code is the right way to go, but we should point users who want this functionality toward newer versions of R.

R/stats-mlm-tidiers.R
```diff
@@ -0,0 +1,88 @@
+#' @templateVar class lm
```
alexpghayes:
```suggestion
#' @templateVar class mlm
```

R/stats-mlm-tidiers.R
```diff
@@ -0,0 +1,88 @@
...
+#' @template title_desc_tidy
+#'
+#' @param x An `lm` object created by [stats::lm()].
```
alexpghayes:
```suggestion
#' @param x An `mlm` object created by [stats::lm()] with a matrix as the response.
```

R/stats-mlm-tidiers.R
```diff
@@ -0,0 +1,88 @@
...
+#' @template param_confint
+#' @template param_quick
+#' @template param_unused_dots
+#'
+#' @evalRd return_tidy(regression = TRUE)
+#'
+#' @details If the linear model is an `mlm` object (multiple linear model),
```
alexpghayes: This is always the case now, so we should update the documentation here.

R/stats-mlm-tidiers.R
```diff
@@ -0,0 +1,88 @@
...
+#'   there is an additional column `response`.
+#'
+#'   If you have missing values in your model data, you may need to refit
+#'   the model with `na.action = na.exclude`.
+#'
+#' @examples
+#'
+#' mod <- lm(cbind(mpg, disp) ~ wt, mtcars)
+#' tidy(mod, conf.int = TRUE)
+#' @aliases lm_tidiers
```
alexpghayes: This line should go since `lm_tidiers` should refer to `tidy.lm()`.
```suggestion
```

R/stats-mlm-tidiers.R
```diff
@@ -0,0 +1,88 @@
...
+#' @export
+#' @seealso [tidy()]
+#' @family lm tidiers
+#' @rdname tidy.mlm
```
alexpghayes: Unnecessary

```suggestion
```

R/stats-mlm-tidiers.R
```diff
@@ -0,0 +1,88 @@
...
+#' @export
+
+tidy.mlm <- function(x,
+                     conf.int = FALSE,
+                     conf.level = .95,
+                     quick = FALSE,
+                     ...) {
+  # if only term and estimate columns are needed
+  if (quick) {
```
alexpghayes: Let's get rid of the `quick` logic, I'm going to strip all of it out for 0.7.0. It's hard to maintain and I don't think anyone uses it.

R/stats-mlm-tidiers.R
```diff
@@ -0,0 +1,88 @@
...
+    co <- stats::coef(x)
+
+    ret <-
+      data.frame(
```
alexpghayes: Can we use `tibble` instead of `data.frame` here?

R/stats-mlm-tidiers.R
```diff
@@ -0,0 +1,88 @@
...
+      )
+
+    return(process_mlm(ret, 
+                      x, 
+                      conf.int = FALSE))
+  }
+
+  # adding other details from summary object
+  s <- summary(x)
+  ret <- tidy.summary.lm(s)
```
alexpghayes: `tidy.summary.lm()` shouldn't exist anymore (?) so you may need to recreate some of that processing here. The code duplication is okay because it'll make things more maintainable / decoupled in the long run.

R/stats-mlm-tidiers.R
```diff
@@ -0,0 +1,88 @@
...
+
+    ret <-
+      data.frame(
+        response = rep(colnames(co), each = nrow(co)),
+        term = rep(rownames(co), times = ncol(co)),
+        estimate = as.numeric(co),
+        stringsAsFactors = FALSE
+      )
+
+    return(process_mlm(ret,
```
alexpghayes: Similarly, I no longer believe in `process_*` helpers. Tidiers should for the most part be a single function that reads linearly. Can you move the `process_mlm()` functionality into the body of `tidy.mlm()`?

R/stats-mlm-tidiers.R
```diff
@@ -0,0 +1,88 @@
...
+#' @export
+glance.mlm <- function(x, ...) {
+  stop(
+    "Glance does not support linear models with multiple responses.",
+    call. = FALSE
+  )
+}
+
+# cleaning up the tidy dataframe
+process_mlm <- function(ret,
```
alexpghayes: See previous comments about moving this into the body of `tidy.mlm()` for an easy to read, linear function.

tests/testthat/test-stats-mlm.R
```diff
@@ -0,0 +1,9 @@
+context("stats-mlm")
+
+fit_mlm <- lm(cbind(mpg, disp) ~ wt, mtcars)
+df_tidy <- tidy(fit_mlm, conf.int = TRUE)
+
+test_that("tidy.mlm works", {
+  testthat::expect_equal(dim(df_tidy), c(4L, 8L))
```
alexpghayes: Can you use `modeltests::check_tidy_output()` and `modeltests::check_dims()` here for consistency with the other tests?

R/stats-mlm-tidiers.R
IndrajeetPatil: The problem is that `process_mlm` function is being used in two different functions and so it will be better to have a separate function to avoid code duplication.

R/stats-mlm-tidiers.R
IndrajeetPatil: The problem is that `process_mlm` function is being used in two different functions and so it will be better to have a separate function to avoid code duplication.

R/stats-mlm-tidiers.R
IndrajeetPatil: I have modified `tidy.summary.lm` to work only with `lm` objects and added a new `tidy.summary.mlm` method for `mlm` objects.

R/stats-lm-tidiers.R
IndrajeetPatil: Hmm, not sure what to do here. Should there be a conditional statement such that if R version is > `3.5`, then it uses `stats::confint` and otherwise uses this custom function? 

I know `R.version` gives `major` (`R.version$major`) and `minor` (`R.version$minor`) version of `R`, but not sure how to get both of them in a string.

R/stats-mlm-tidiers.R
alexpghayes: When we get rid of the `quick` argument it only gets called in one place.

R/stats-lm-tidiers.R
alexpghayes: Let's `tryCatch()` the `confint()` call that is currently in `process_mlm()` then, with a warning about R versions when that fails.

R/stats-lm-tidiers.R
alexpghayes: This change didn't make it through?

R/stats-mlm-tidiers.R
IndrajeetPatil: Okay, I will postpone refactoring to a later date then - when this argument is actually retired.

R/stats-mlm-tidiers.R
IndrajeetPatil: Awesome! Happy to get rid of this. Didn't realize we were deprecating this in `0.7.0`.

IndrajeetPatil:
@alexpghayes Done!

R/stats-mlm-tidiers.R
```diff
@@ -0,0 +1,99 @@
...
+                     conf.int = FALSE,
+                     conf.level = .95,
+                     ...) {
+
+  # adding other details from summary object
+  s <- summary(x)
+  ret <- tidy.summary.mlm(s)
+
+  # adding confidence intervals
+  if (isTRUE(conf.int)) {
```
alexpghayes:
```suggestion
  if (conf.int) {
```

R/stats-mlm-tidiers.R
```diff
@@ -0,0 +1,99 @@
...
+
+    # S3 method for computing confidence intervals for `mlm` objects was
+    # introduced in R 3.5
+    CI <- tryCatch(
```
alexpghayes: This isn't quite what I meant. This tries to get a CI and doesn't add the CI if things break. This makes the tidier sort of magical, in that the user doesn't know what columns they are going to get back. If the user requests tidiers and can't get them, we should throw an informative error, rather than returning the standard columns other than `conf.low` and `conf.high`.

R/stats-mlm-tidiers.R
```diff
@@ -0,0 +1,99 @@
...
+      expr = stats::confint(x, level = conf.level),
+      error = function(x) {
+        NULL
+      }
+    )
+
+    # if R version is prior to 3.5, use the custom function
+    if (is.null(CI)) {
```
alexpghayes: We deleted `confint.mlm`? So this shouldn't work?

R/stats-mlm-tidiers.R
```diff
@@ -0,0 +1,99 @@
...
+      CI <- confint.mlm(x, level = conf.level)
+    }
+
+    colnames(CI) <- c("conf.low", "conf.high")
+    ret <- cbind(ret, unrowname(CI))
```
alexpghayes: I prefer to use `bind_cols` and `tibble` so that we don't need to use `cbind()` and `unrowname()`, tools designed mostly for data frames.

R/stats-mlm-tidiers.R
```diff
@@ -0,0 +1,99 @@
...
+  }
+
+  tibble::as_tibble(ret)
```
alexpghayes: `as_tibble()` is imported so no need to namespace it.

```suggestion
  as_tibble(ret)
```

R/stats-mlm-tidiers.R
```diff
@@ -0,0 +1,99 @@
...
+}
+
+# mlm objects subclass lm objects so this gives a better error than
+# letting augment.lm() fail
+#' @include null-and-default-tidiers.R
+augment.mlm <- augment.default
+
+#' @export
+glance.mlm <- function(x, ...) {
```
alexpghayes: It now strikes me as odd that we have different error strategies for `augment` and `glance`. Let's do 

```r
glance.mlm <- glance.default
```

for consistency.

R/stats-mlm-tidiers.R
```diff
@@ -0,0 +1,99 @@
...
+  stop(
+    "Glance does not support linear models with multiple responses.",
+    call. = FALSE
+  )
+}
+
+#' @export
+tidy.summary.mlm <- function(x, ...) {
```
alexpghayes: Can we move this logic into `tidy()` itself? A single linear function is easier to understand, and more safe to modify.

R/stats-mlm-tidiers.R
```diff
@@ -0,0 +1,99 @@
...
+  ret <- purrr::map_df(co, fix_data_frame, nn[1:ncol(co[[1]])],
+    .id = "response"
+  )
+  ret$response <- stringr::str_replace(ret$response, "Response ", "")
+
+  as_tibble(ret)
+}
+
+# compute confidence intervals for mlm object.
+confint.mlm <- function(object, level = 0.95, ...) {
```
alexpghayes: Oh I see it made it back down here. I'm happy to keep this, or alternatively to just let this be an error. If you want to support R < 3.5, instead of check if the CI is `NULL`, this should just be directly in the catch of `tryCatch`.

R/stats-lm-tidiers.R
IndrajeetPatil: Huh. Sorry, missed that I had also changed `lm` tidier files.

R/stats-mlm-tidiers.R
IndrajeetPatil: No, we still have it. The current strategy is-
First try to use the `stats::confint` method and if it fails, use the custom `S3` method in the package.  

It makes sense to delete `confint.mlm` only when `broom` no longer supports `R` version `3.4.0`

R/stats-mlm-tidiers.R
IndrajeetPatil: No, this always returns CI, either using `stats::confint` or using the custom `confint` method in `broom` (which has reintroduced to support `3.4.0`).

R/stats-mlm-tidiers.R
IndrajeetPatil: Hmm, unfortunately, I don't have much experience with `tryCatch` so I am not sure how to refactor this in a way that satisfied your requirements.

IndrajeetPatil:
@alexpghayes I've addressed your new comments.

R/stats-mlm-tidiers.R
alexpghayes: This criticism is valid when you want to use `all.equal()` but not valid in general. `if(isTRUE(logical_var))` is redundant. We ask the user to supply a logical to `conf.int`, so it's reasonable to operate on the assumption they've done this. If we want to valid that `conf.int` is a logical and not `NA` or `NaN` or `NULL`, we could do that, but we should separate that logical out into distinct input validation infrastructure. At some point, we should add input validation to tidiers. That's currently a low priority, however.

R/stats-mlm-tidiers.R
alexpghayes: No problem, I'll play around with it.

IndrajeetPatil:
Okay, I've accepted your commit.

alexpghayes:
Great! Thanks!

PR #665, opened 2019-03-20
Use Circle CI and misc cleanup

IndrajeetPatil:
In addition to the `AppVeyor` and `Travis`, `CircleCI` is another popular continuous integration platform service. 

(For more, see: https://yutani.rbind.io/post/2017-10-18-circleci/)


alexpghayes:
Thanks! Haven't used Circle CI before -- do I need to do any set up?

alexpghayes:
Also I don't think we need the Docker image with all the database stuff since we're not developing database stuff in `broom`. Is there a more lightweight docker image we could use?

Issue #628, opened 2019-03-06
Allow tidy.lavaan() to take `quick = TRUE`, or make this an error


alexpghayes:
Original reported in #558.

```
set.seed(123)
library(lavaan)
#> This is lavaan 0.6-3
#> lavaan is BETA software! Please report any bugs.

# model
cfa.fit <- cfa('F =~ x1 + x2 + x3 + x4',
               data = HolzingerSwineford1939, group = "school")

# tidieer
broom::tidy(
  x = cfa.fit,
  conf.int = TRUE,
  conf.level = 0.99,
  quick = TRUE
)
#> Error in lavaan::parameterEstimates(x, ci = conf.int, level = conf.level, : unused argument (quick = TRUE)
```

PR #671, opened 2019-04-02
Fix tidy.lavaan() bug to allow quick = TRUE (#628)

x249wang:


alexpghayes:
I appreciate this PR and am going to merge! Just an FYI, I'm become convinced that the `quick` behavior is not desirable in general, and plan to remove all `quick` functionality at some unspecificied point in the future.

PR #674, opened 2019-04-04
metafor::rma tidiers - work in progress

softloud:
These tidiers are still _work in progress_ with @malcolmbarrett. 

Will ping you @alexpghayes, when we're ready for you to take a look.

## resources

- `broom::` helpfully has pointers on what to do for a [pr](https://cran.r-project.org/web/packages/broom/vignettes/adding-tidiers.html)
- `broom::` [test examples](https://github.com/tidymodels/broom/blob/master/tests/testthat/test-mass-rlm.R)
- [10 steps to being a tidyverse contributor](https://thisisnic.github.io/2018/11/28/ten-steps-to-becoming-a-tidyverse-contributor/)
- [modeltests](https://github.com/softloud/metabroom/issues/15#issuecomment-480129749) documentation
- this advice from @alexpghayes
- the [original thread](https://github.com/softloud/metabroom/issues/15) discussing the tests

<blockquote class="twitter-tweet" data-lang="en"><p lang="en" dir="ltr">I recommend starting with strict = FALSE for all the tests. Then do tidy() and glance() with strict = TRUE. Only then do augment() with strict = TRUE. augment() is astonishingly painful to get right.</p>&mdash; alex hayes (@alexpghayes) <a href="https://twitter.com/alexpghayes/status/1113992682358153216?ref_src=twsrc%5Etfw">April 5, 2019</a></blockquote>

## tasks

- [x] convert Malcolm's script to tests
- [x] figure out why methods aren't being picked up
- [x] mine [tidyverse contributor](https://thisisnic.github.io/2018/11/28/ten-steps-to-becoming-a-tidyverse-contributor/) post for git tips
- [x] add us as contributors
- [x] Malcolm to add his ORCID


softloud:
Tests are failing because the methods aren't recognised. 

My new attained [fluffy duck](http://cantabile.rbind.io/posts/2019-03-28-quack-quack-said-the-duck/) understanding of S3 is failing me. I think there's stuff in the `roxygen` documentation in the other tidier files that will fix this? 

Is there a guide or template for this documentation, @alexpghayes?

R/rma-tidiers.R
```diff
@@ -0,0 +1,222 @@
...
+#' These methods tidy the results of meta-analysis objects from the metafor package
+#'
+#' @param x An `rma` created by the `metafor` package.
+#' @inheritParams tidy.lm
+#' @param include_studies Logical. Should individual studies be included in the
+#'    output?
+#' @param ... Additional arguments
+#' @param measure Measure type. See [metafor::rma()]
+#'
+#' @return A `tibble`
```
alexpghayes: Take a look at https://github.com/tidymodels/broom/blob/master/R/betareg-tidiers.R#L8 for an example of how to document the tidier return. There are three separate functions `return_tidy()`, ..., `return_augment()` to look at. Also I definitely need to document how to use these better.

R/rma-tidiers.R
```diff
@@ -0,0 +1,222 @@
...
+#'     ci = cpos,
+#'     di = cneg,
+#'     data = dat.bcg
+#'   )
+#' 
+#' meta_analysis <- rma(yi, vi, data = df, method = "EB")
+#'
+#' tidy(meta_analysis)
+#'
+#' @rdname tidiers
```
alexpghayes:
```suggestion
#' @rdname metafor_tidiers
```

R/rma-tidiers.R
```diff
@@ -0,0 +1,222 @@
...
+#' 
+tidy.rma <- function(x, conf.int = FALSE, conf.level = 0.95, exponentiate = FALSE,
+                     include_studies = TRUE, measure = "GEN", ...) {
```
alexpghayes: Can use `measure = c("GEN", "other_option1", "other_option2")` etc to make the set of all options clear? And then use `rlang::arg_match()` to validate this argument?

R/rma-tidiers.R
```diff
@@ -0,0 +1,222 @@
...
+  
+  estimates <- metafor::escalc(yi = x$yi.f, vi = x$vi.f, measure = measure) %>%
+    summary(level = conf.level * 100) %>% 
+    as.data.frame(stringsAsFactors = FALSE)
+  
+  estimates <- cbind(x$slab, "study", estimates[, c("yi", "sei", "zi")], NA,
```
alexpghayes: Can I ask why the `NA` here? Also, would you consider using `bind_cols()` instead?

R/rma-tidiers.R
```diff
@@ -0,0 +1,222 @@
...
+    p.value.qm = x$QMp,
+    fit_stats
+  ) %>%
+    # get rid of null values
+    purrr::discard(is.null) %>%
+    # don't include multivariate model stats
+    purrr::discard(~length(.x) >= 2) %>%
+    # change to tibble with correct column and row names
+    as.data.frame() %>%
+    tibble::as_tibble() %>%
```
alexpghayes: I don't think `tibble::remove_rownames()` is necessary if you use `tibble::as_tibble()`?

R/rma-tidiers.R
```diff
@@ -0,0 +1,222 @@
...
+#'     data = dat.bcg
+#'   )
+#' 
+#' meta_analysis <- rma(yi, vi, data = df, method = "EB")
+#'
+#' augment(meta_analysis)
+#'
+#' @rdname augmenters
+augment.rma <- function(x, ...) {
+  blup0 <- purrr::possibly(metafor::blup, NULL)
```
alexpghayes: Why is it necessary to use `possibly` here? Generally, this makes me sort of nervous because it's very hard to reason about which columns will actually be returned.

R/rma-tidiers.R
```diff
@@ -0,0 +1,222 @@
...
+    .rownames = x$slab,
+    y,
+    pred,
+    .resid = res
+  )
+  
+  ret <- tibble::as_tibble(ret) %>%
+    tibble::remove_rownames()
+  
+  if (all(ret$.rownames == seq_along(ret$.rownames))) {
```
alexpghayes: I think there's a `has_rownames()` helper somewhere that you can use here?

tests/testthat/test-rma.R
```diff
@@ -0,0 +1,207 @@
+context("rma")
```
alexpghayes: Would you mind structuring the tests based on method? This is what the other tests files do. For an example you could look at: https://github.com/tidymodels/broom/blob/master/tests/testthat/test-aer.R#L33

tests/testthat/test-rma.R
```diff
@@ -0,0 +1,207 @@
...
+    di = di,
+    data = dat,
+    measure = "RD",
+    digits = 3,
+    level = 90
+  )
+
+test_that(("returns tibble for MH"), {
+  check_tidy_output(tidy(res.MH))
+  # check_augment_outputs(glance(res.MH))
```
alexpghayes: Commented out tests can be confusing. I would either note why this is commented out or delete the code.

R/rma-tidiers.R
malcolmbarrett: You're right! These are leftover from when I first wrote these, which was prior to broom using tibble

R/rma-tidiers.R
malcolmbarrett: Yeah, I get that: basically, metafor handles also this from a single `predict` method for all models; some use `blup()` and some use `predict.rma`. This approach gets the predictions from `blup` when that's what is used, `predict` otherwise; the two are mutually exclusive (blup will always return `NULL` for those that don't use it). I don't really get why it's set up like that, but it seems to be robust to model type

R/rma-tidiers.R
malcolmbarrett: I'm going to change the approach in these lines to make it a little clearer what is happening because the rownames aren't coming from the usual place in this model

tests/testthat/test-rma.R
malcolmbarrett: oh, I didn't notice this myself. `check_augment_outputs` isn't a function in `modeltests` yet, right? we can replace with a simpler check for tibble output in the augment functions

malcolmbarrett: Or should we use `modeltests::check_augment_function`?

R/rma-tidiers.R
malcolmbarrett: `NA` is for the p values for the individual studies, which aren't in the model. Only the summary estimate returns a p value. I don't mind switching to `bind_cols()`, but I'll have to change a few things since it's strict about all the objects having the same length

R/rma-tidiers.R
```diff
@@ -1,24 +1,34 @@
...
-#' @param include_studies logical. Should individual studies be included in the
+#' These methods tidy the results of meta-analysis objects from the metafor package
+#'
+#' @param x An `rma` created by the `metafor` package.
+#' @inheritParams tidy.lm
+#' @param include_studies Logical. Should individual studies be included in the
 #'    output?
-#' @param ... additional arguments
-#' @param measure measure type. See [metafor::rma()]
+#' @param ... Additional arguments
```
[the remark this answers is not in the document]
malcolmbarrett: Ah, they actually aren't used but need to be there to match the generic. I seem to recall some methods in broom saying something like `Additional arguments (not used)`?

R/rma-tidiers.R
malcolmbarrett: The list is very long (60!) and probably unwieldy for this? I might link to `metafor::escalc()` instead, which lists all of them. `metafor::escalc()`does check for validity internally

tests/testthat/test-rma.R
malcolmbarrett: After exploring `modeltests::check_augment_function` a bit, I see I'd run into the problem with the `data` argument even when `strict = FALSE`

R/rma-tidiers.R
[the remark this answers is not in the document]
alexpghayes: If they aren't getting used the standard way to document them is with `@template param_unused_dots`

R/rma-tidiers.R
alexpghayes: If we need to keep all these `possibly`s that's okay, but comments documenting when you get `NULL` and when you get something would be really helpful for maintenance down the line.

tests/testthat/test-rma.R
alexpghayes: The augment checks currently are designed to make sure prediction on newdata doesn't blow up, more or less. I would use `check_tidy_output()` and `check_glance_outputs()` and then write some simple custom checks in for the `augment.rma()` method.

malcolmbarrett:
Changes implemented, @alexpghayes!

R/rma-tidiers.R
```diff
@@ -0,0 +1,288 @@
+#' @templateVar class rma
+#' @template title_desc_tidy
+#' 
+#' @param x An `rma` created by the `metafor` package.
+#' @inheritParams tidy.lm
+#' @param include_studies Logical. Should individual studies be included in the
+#'    output?
```
alexpghayes:
```suggestion
#'    output? Defaults to `TRUE`.
```

R/rma-tidiers.R
```diff
@@ -0,0 +1,288 @@
+#' @templateVar class rma
+#' @template title_desc_tidy
+#' 
+#' @param x An `rma` created by the `metafor` package.
```
alexpghayes: I would link to all the function that create rma objects here. That is, something along the lines of:

> #' @param x An `rma` such as those created by [metafor::function1()], [metafor::function2()] or ....

R/rma-tidiers.R
```diff
@@ -0,0 +1,288 @@
...
+  if (exponentiate) {
+    results <-  dplyr::mutate_at(results, dplyr::vars(estimate, conf.low, conf.high), exp)
+  }
+  
+  if (!conf.int) {
+    results <- dplyr::select(results, -conf.low, -conf.high)
+  }
+  
+  # remove extra model data from study names
+  attributes(results$study) <- NULL
```
alexpghayes: Why is this necessary?

softloud:
Wow, @malcolmbarrett - this is amazing! Taking a look to see what I can do now.

alexpghayes:
This is looking fantastic. Left three more short comments.

R/rma-tidiers.R
malcolmbarrett: Hm, metafor stores a lot of extra information in these vectors which caused problems in working with data frames, but I suspect being stricter while assembling `tibbles` and using `dplyr::bind_*()` already stripped these and now it's no longer necessary.

malcolmbarrett:
Thanks, @alexpghayes and @softloud. Addressed!

alexpghayes:
Fantastic work! Thrilled to have this in `broom`, thank you so much!

malcolmbarrett:
Yess, thanks for all your help, @alexpghayes!

PR #680, opened 2019-04-07
Updates for 0.5.2 and remove quick argument

alexpghayes:
- [ ] Remove quick from allowed arguments in `modeltests`


alexpghayes:
I don't know why git is angry so I'm doing a hacky workaround instead of this. I apologize to anyone who values git history.

Issue #115, opened 2016-03-07
t.test with factors should use factor labels instead of estimate1/2


jankatins:
Example:

``` R
df <- data.frame(a=1:8, b=factor(c(0,1,0,1,0,1,0,1), levels=c(0,1), labels=c("a", "b")) )
print(df %>% t.test(a~b, data = .))
print(df %>% t.test(a~b, data = .) %>% tidy())
```

results in 

```
    Welch Two Sample t-test

data:  a by b
t = -0.54772, df = 6, p-value = 0.6036
alternative hypothesis: true difference in means is not equal to 0
95 percent confidence interval:
 -5.467429  3.467429
sample estimates:
mean in group a mean in group b      ######### <- factor labels
              4               5 

  estimate estimate1 estimate2  statistic   p.value parameter  conf.low     ####### <- no factor labels :-(
1       -1         4         5 -0.5477226 0.6036451         6 -5.467429
  conf.high
1  3.467429
```

alexpghayes:
I think this makes using a single t-test object nicer. It would make working with lists of t-tests much harder, however -- the following would break due to differing column names:

```r
purrr::map_dfr(list_of_tests, tidy)
```

alexpghayes:
I like this! I think having `x` and `y` as group names isn't ideal, but it's the best workaround and we can document it carefully.

nutterb:
Included in #326

Issue #31, opened 2015-03-02
Retaining statistics names


crsh:
I'm working on an [R package of my own](https://github.com/crsh/papaja), which will provide functions to knit scientific manuscripts.

To this end, I'm building convenience functions that assemble text strings form analysis objects, such as `htest` or `summary.lm` etc. I am considering to use `broom` to tidy those objects. The only thing that is currently keeping me from using your package is that fact that when I tidy objects I loose all information about what the `estimates` actually are (differences of means, means of differences, correlation coefficients, etc.)---the same is, of course, true for other columns of the tidied output.

Are there any plans to add this information to tidied data.frames? I think retaining this information would be helpful for other purposes and programming literacy in general.

An unobtrusive way of doing this would be to simply add attributes to the data.frame from the original object (adding a row to the data.frame would be another possibility). I've thrown something together for an object from `t.test()` to illustrate what I mean (this should generalize to other `htest` objects with some minor adaptations):

    > t_test <- t.test(extra ~ group, data = sleep)    
    > tidy_t_test <- tidy(t_test)
    > tidy_t_test
    
      estimate estimate1 estimate2 statistic    p.value parameter  conf.low conf.high
    1    -1.58      0.75      2.33 -1.860813 0.07939414  17.77647 -3.365483 0.2054832
    
    > vars <- lapply(t_test, attr, "names")
    > vars <- vars[!unlist(lapply(vars, is.null))]
    > conf_level <- attr(t_test$conf.int, "conf.level") * 100
    > conf_levels <- paste(c((100 - conf_level) / 2, 100 - (100 - conf_level) / 2), "%")
    
    > attr(tidy_t_test, "vars") <- c(vars$null.value, vars$estimate, vars$statistic, "p.value", vars$parameter, conf_levels)
    > str(tidy_t_test)
    
    'data.frame':  1 obs. of  8 variables:
    $ estimate : num -1.58
    $ estimate1: num 0.75
    $ estimate2: num 2.33
    $ statistic: num -1.86
    $ p.value  : num 0.0794
    $ parameter: num 17.8
    $ conf.low : num -3.37
    $ conf.high: num 0.205
    - attr(*, "vars")= chr  "difference in means" "mean in group 1" "mean in group 2" "t" ...
    
    > attr(tidy_t_test, "vars")
    [1] "difference in means" "mean in group 1" "mean in group 2" "t" "p.value" "df" "2.5 %"              
    [8] "97.5 %"

[...]

PR #696, opened 2019-04-19
Update stats-htest-tidiers.R

rsbivand:
Check whether ret$estimate is named before assigning names, if named do not overwrite names (similar logic to checking for parameter names in the same method).

### DIFF
--- a/R/stats-htest-tidiers.R
+++ b/R/stats-htest-tidiers.R
@@ -47,7 +47,9 @@ tidy.htest <- function(x, ...) {
 
   # estimate may have multiple values
   if (length(ret$estimate) > 1) {
-    names(ret$estimate) <- paste0("estimate", seq_along(ret$estimate))
+    if (is.null(names(ret$estimate))) {
+      names(ret$estimate) <- paste0("estimate", seq_along(ret$estimate))
+    }
     ret <- c(ret$estimate, ret)
     ret$estimate <- NULL
 


rsbivand:
@alexpghayes Please motivate your rejection - I do not know of, or was asked about, the unnumbered issue to which you refer.

alexpghayes:
Right. Apologies for the hasty decision making -- I've been try to cram a lot of maintenance into a short time period. Two relevant issues are #31 and #115. Revisiting #115 it looks like we only partially came to a conclusion on what to do and then the implementation got absorbed in the #326, which is a behemoth PR that needs a lot of work.

The main concern I recall with informative names for the means was that it would make `map*`-ing difficult. One option is to add an argument that controls whether you get group names or `estimate1` and `estimate2` , and default to group names. What do you think about this?

rsbivand:
Thanks for getting back to me. I am not convinced by map/group arguments, and without a reprex say using one of the following standard tests would find it hard to think of a relevant context. By dropping names added for good reasons by function authors, potentially important information is lost, I think.

Some (legacy) examples of named estimate components of htest objects - most of these have meaningful names, often recording how the test was applied:

```
library(stats)
names(ansari.test(rnorm(10), rnorm(10, 0, 2), conf.int = TRUE)$estimate)
names(binom.test(c(682, 243), p = 3/4)$estimate)
names(cor.test(rnorm(5), rnorm(5), method = "kendall", alternative = "greater")$estimate)
TeaTasting <- matrix(c(3, 1, 1, 3),  nrow = 2, dimnames = list(Guess = c("Milk", "Tea"), Truth = c("Milk", "Tea")))
names(fisher.test(TeaTasting, alternative = "greater")$estimate)
Rabbits <- array(c(0, 0, 6, 5, 3, 0, 3, 6, 6, 2, 0, 4, 5, 6, 1, 0, 2, 5, 0, 0), dim = c(2, 2, 5),        dimnames = list(Delay = c("None", "1.5h"), Response = c("Cured", "Died"), Penicillin.Level = c("1/8", "1/4", "1/2", "1", "4")))
names(mantelhaen.test(Rabbits)$estimate)
names(poisson.test(137, 24.19893)$estimate)
smokers  <- c( 83, 90, 129, 70 )
patients <- c( 86, 93, 136, 82 )
names(prop.test(smokers, patients)$estimate)
names(t.test(extra ~ group, data = sleep)$estimate)
x <- rnorm(50, mean = 0, sd = 2)
y <- rnorm(30, mean = 1, sd = 1)
names(var.test(x, y)$estimate)
x <- c(1.83,  0.50,  1.62,  2.48, 1.68, 1.88, 1.55, 3.06, 1.30)
y <- c(0.878, 0.647, 0.598, 2.05, 1.06, 1.29, 1.06, 3.14, 1.29)
names(wilcox.test(x, y, paired = TRUE, alternative = "greater", conf.int=TRUE)$estimate)
```

alexpghayes:
Cycling back to this. Apologies for the uncommunicative reversion. I believe that this PR ended up breaking something that I reverted in a last minute fix for CRAN. My lack of responsiveness here is a largely a function of my limited capacity for maintenance. I do agree that this is a good idea and also that the former goal of preserving the ability to `map` isn't that important. 

I think a good next step here would be to make re-institute this change and run `revdepcheck`, since I vaguely recall that's where issues emerged last time. I can put this on my own todo list and it will happen eventually, but a new PR with `revdepcheck` results would speed things up if you're looking for this functionality faster.

PR #705, opened 2019-05-23
Convert zoo tidier to respect column name

angusmoore:
Zoo column names can have non-`make.name`s names, e.g., they can have spaces in them.

The tidier for zoos was (implicitly) running `make.names`, so zoos with non-standard column names would get different names in series column once tidied than what they had as a zoo.

E.g. for this zoo:
```
z.index <- zoo::as.Date(sample(12450:12500, 10))
z.data <- matrix(rnorm(30), ncol = 3)
colnames(z.data) <- c("Not dataframe", "(compatible", "names -")
z <- zoo::zoo(z.data, z.index)
```

When it gets tidied, the second statement isn't true:
```
td <- tidy(z)
all(unique(td$series) %in% colnames(z.data))
```

This PR changes the behaviour of the zoo tidier so that the second statement _is_ true. I think that's the correct behaviour, but I am not fully across the arguments for and against `make.names`, so am happy if there is a reason to keep the existing behaviour.


alexpghayes:
Nice catch! I agree, I think this is the correct behavior. Will you address yourself as a contributor in `DESCRIPTION` and then I'll merge?

angusmoore:
Thanks, have added myself, and resolved conflicts. The appveyor failure looks spurious: caused by a failure to install curl. Unfortunately, I don't think I can restart the build.

alexpghayes:
Yep, looks good! Thank you!

Issue #714, opened 2019-06-13
confint method for `geeglm` objects not working well


iago-pssjd:
I tried to use the method [confint.geeglm](https://www.rdocumentation.org/packages/broom/versions/0.4.5/topics/confint.geeglm), but I got the following:

```
# creating a dataframe
set.seed(123)
library(geepack)
data(state)
ds <- data.frame(state.region, state.x77)

# model
geefit <-
  geepack::geeglm(
    formula = Income ~ Frost + Murder,
    id = state.region,
    data = ds,
    family = gaussian,
    corstr = "exchangeable"
  )

# confint.geeglm method
confint(geefit)
#> Waiting for profiling to be done...
#> Error in `[.data.frame`(summ$coefficients, , "Std. Error", drop = FALSE) : 
#>   undefined columns selected

# But it works if it is specified
broom:::confint.geeglm(geefit)
#>                  lwr      upr
#> (Intercept) 3607.596 5204.790
#> Frost         -2.723    6.097
#> Murder       -84.167   38.838
```

IndrajeetPatil:
I can't reproduce this error. Are you by any chance using the `CRAN` version of `broom`? I'd recommend updating to the development version on `GitHub` which has taken care of a lot of known issues with some of the tidiers.

``` r
# creating a dataframe
set.seed(123)
library(geepack)
data(state)
ds <- data.frame(state.region, state.x77)

# model
geefit <-
  geepack::geeglm(
    formula = Income ~ Frost + Murder,
    id = state.region,
    data = ds,
    family = gaussian,
    corstr = "exchangeable"
  )

# tidy output with confidence intervals
broom::tidy(x = geefit, conf.int = TRUE)
#> # A tibble: 3 x 7
#>   term        estimate std.error statistic p.value conf.low conf.high
#>   <chr>          <dbl>     <dbl>     <dbl>   <dbl>    <dbl>     <dbl>
#> 1 (Intercept)  4406.      407.     117.      0      3608.     5205.  
#> 2 Frost           1.69      2.25     0.562   0.453    -2.72      6.10
#> 3 Murder        -22.7      31.4      0.522   0.470   -84.2      38.8

# with the helper function
broom::confint_tidy(geefit)
#> # A tibble: 3 x 2
#>   conf.low conf.high
#>      <dbl>     <dbl>
#> 1  3608.     5205.  
#> 2    -2.72      6.10
#> 3   -84.2      38.8
```

[...]

PR #716, opened 2019-06-18
Export broom:::confint.geeglm to solve issue#714

iago-pssjd:
Export `broom:::confint.geeglm` to solve issue #714


alexpghayes:
Duplicate of #731, closing for the same reasons.

Issue #672, opened 2019-04-03
Make sure exponentiate arguments default to FALSE


mattpollock:
exponentiate | Logical indicating whether or not to exponentiate the the coefficient estimates. This is typical for logistic and multinomial regressions, but a bad idea if there is no log or logit link. Defaults to FALSE.
-- | --

However, under *Usage*
```
tidy(x, conf.int = FALSE, conf.level = 0.95,
  exponentiate = TRUE, ...)
```

FWIW I think the documented default to `FALSE` is better - then the output matches the `summary` output and since there is not indication in the output tibble showing the difference I think a `TRUE` default is more likely to lead to errors where people don't realize that the estimates are already exponentiated and to exponentiate again.

alexpghayes:
Nice catch. I'll add a test to make sure everything using `exponentiate` defaults to FALSE.

alexpghayes:
Tests should now give off errors from https://github.com/alexpghayes/modeltests/blob/master/R/check_arguments.R#L61. Fixing those failures should result in consistent documentation and behavior.

dylanjm:
Has a PR been submitted for this issue? I can't find one and it seems a lot of tests already incorporate `check_arguments()`.

alexpghayes:
Got it in 9b8fb69344cfdbaca4645c7120e4498dcce34fd5!

PR #725, opened 2019-07-09
Changed nnet-tidiers to default exponentiate = FALSE refs #672

dylanjm:
refs #672

This is the only file that seemed to error out due to testing.


alexpghayes:
Looks like I got this in 9b8fb69344cfdbaca4645c7120e4498dcce34fd5, which just took forever to merge. Thanks for the PR!

PR #740, opened 2019-08-13
Tidy orcutt issue 734

### LINKED ISSUE #734: tidy.orcutt() errors on confidence intervals and exponentiation
Bug originally reported in #712 and migrated to this issue

```r
library(orcutt)
set.seed(123)
reg <- stats::lm(formula = mpg ~ wt + qsec + disp, data = mtcars)
co <- orcutt::cochrane.orcutt(reg)

# plot
broom::tidy(
  x = co,
  title = "Cochrane-Orcutt estimation",
  conf.int = TRUE,
  exponentiate = FALSE
)
#> Warning in tidy.orcutt(x = co, title = "Cochrane-Orcutt estimation",
#> conf.int = TRUE, : deal with tidy.orcutt conf.int nonsense
#> Warning in if (exponentiate) {: the condition has length > 1 and only the
#> first element will be used
#> Error in if (exponentiate) {: argument is not interpretable as logical
```

A PR to solve this issue should feature regression tests. Honestly I'm not sure what's going on with the `orcutt` tidiers, but it's likely that it isn't well designed and many arguments can just get dropped IIRC.

jyuu:
Fixes #734


R/orcutt-tidiers.R
```diff
@@ -28,12 +28,22 @@
...
+  s <- summary(x)
+  ret <- tidy.summary.orcutt(s)
+  
+  process_orcutt(ret, x, exponentiate = exponentiate)
+}
 
-  rlang::exec(tidy.lm, x, !!!dots)
+#' @rdname orcutt
+#' @export
+tidy.summary.orcutt <- function(x, ...) {
```
alexpghayes: As a general rule I prefer to have a single `tidy()` method that works on the original object, without an extra `tidy()` method for the summary. Tidying a summary feels like double-tidying to me, and I don't really know how to think about it.

R/orcutt-tidiers.R
```diff
@@ -70,3 +80,26 @@ glance.orcutt <- function(x, ...) {
     nobs = stats::nobs(x)
   )
 }
+
+
+process_orcutt <- function(ret,
```
alexpghayes: Can we move this logic directly into `tidy.orcutt()`? I think the `process_*()` functions are unnecessary and make it harder to following the tidying logic. i.e. I would turn this PR into a single function `tidy.orcutt()` rather than three separate functions. Also, I don't think the exponentiation logic is unnecessary here -- that's typically only used for GLM/logistic type tidiers that get shoved through `tidy.lm()`.

jyuu: done!

R/orcutt-tidiers.R
```diff
@@ -29,11 +29,11 @@
...
 tidy.orcutt <- function(x, ...) {
-  warning("deal with tidy.orcutt conf.int nonsense")
-  dots <- enquos(...)
-  dots$conf.int <- FALSE
-
-  rlang::exec(tidy.lm, x, !!!dots)
+  s <- summary(x)
+  co <- stats::coef(s)
+  nn <- c("estimate", "std.error", "statistic", "p.value")
+  ret <- fix_data_frame(co, nn[1:ncol(co)])
```
alexpghayes: Much cleaner, I love it! I'm also trying to move away from using `fix_data_frame()`. I think if you use `ret <- as_tibble(..., rownames = "term")` and then set colnames on `ret` we could be able to get rid of the `fix_data_frame()` call.

R/orcutt-tidiers.R
```diff
@@ -29,11 +29,11 @@
...
-  warning("deal with tidy.orcutt conf.int nonsense")
-  dots <- enquos(...)
-  dots$conf.int <- FALSE
-
-  rlang::exec(tidy.lm, x, !!!dots)
+  s <- summary(x)
+  co <- stats::coef(s)
+  ret <- as_tibble(co, rownames = "term")
+  names(ret) <- c("term", "estimate", "std.error", "statistic", "p.value")
+  return(ret)
```
alexpghayes: In `broom` we typically use implicit returns.

```suggestion
  ret
```

PR #770, opened 2019-10-17
Fix issue 765

jcfisher:
This edit allows `tidy` to parse the predictor variables in a `survfit` model.  It works by pulling the model terms from the "call" element of the survfit object, looking for those terms in the text string, and extracting the string between the terms.

I'm a GitHub newbie, so please let me know if there's anything I can do to improve this contribution.  And thanks for putting together this package!

Jake


alexpghayes:
Hey, thanks for this PR! Can you:

- [ ] write some tests to confirm that your code works
- [ ] update `NEWS` to describe your changes
- [ ] add yourself as a contributor in `DESCRIPTION`

and then I'll take a closer look!

jcfisher:
Okay, I've taken care of those things.  This should be ready for you to look over, whenever you get a chance.  

I set this up to parse the strata string by looking for `=` and `,`, which is how `survival` pastes together strata names and values.  The approach that I put together is pretty conservative -- if it finds an extra `=` or `,`, it produces a warning and doesn't create parsed columns.  

The idea was that column names and column values can both have `=` or `,` in them, and there's no way to find out what the "true" values should be without looking at another object in the environment.  So for those edge cases, I thought it would make more sense for users to figure out their own parsing schemes, rather than to have a fragile parser that they have to troubleshoot.

Please let me know if this looks good or if there are other changes you'd like to see.  Thanks!

Issue #778, opened 2019-11-04
.resid column has opposite signs for (at least) augment.lm()


carlislerainey:
This is a bug report for the development version of broom as of
2019-10-04 (`master`, SHA 01181bb85f9db3e3bd727033564008d1186ee65a).

# The Error

Load the required packages.

``` r
library(tidyverse)
library(kableExtra)

# use the development version of broom
devtools::install_github("tidymodels/broom", 
                         ref = "01181bb85f9db3e3bd727033564008d1186ee65a")
library(broom)
```

First, simulate some fake data and fit a model to that dataset.

``` r
set.seed(1234)

x <- seq(-1, 1, by = 0.5)
y <- 0.5*x + rnorm(length(x))

df <- tibble(x, y)

fit <- lm(y ~ x, data = df)
```

Use broom\!

``` r
au <- augment(fit) 
```

Plot the residuals, which should be the observed values minus the fitted values.

``` r
ggplot(au, aes(x = x, y = y, xend = x, yend = .fitted)) + 
  geom_segment() + 
  geom_smooth(method = "lm", se = FALSE) +  
  geom_point() + 
  geom_label(aes(x = x, 
                       y = (.fitted + y)/2, 
                       label = round(.resid, 2), 
                       color = factor(sign(.resid))),
                   size = 2)
```

![unnamed-chunk-4-1](https://user-images.githubusercontent.com/1969854/68124600-c5b96580-fedd-11e9-8151-0840b70ff63d.png)

Inspect the residuals. Notice that the residuals have the wrong signs.

``` r
au %>%
  select(y, x, .fitted, .resid) %>%
  mutate(.resid_manual = y - .fitted) %>%
  kable(format = "markdown", digits = 2)
```

|      y |     x | .fitted | .resid | .resid\_manual |
| -----: | ----: | ------: | -----: | -------------: |
| \-1.71 | \-1.0 |  \-0.98 |   0.72 |         \-0.72 |
|   0.03 | \-0.5 |  \-0.67 | \-0.69 |           0.69 |
|   1.08 |   0.0 |  \-0.35 | \-1.44 |           1.44 |
| \-2.10 |   0.5 |  \-0.04 |   2.06 |         \-2.06 |
|   0.93 |   1.0 |    0.28 | \-0.65 |           0.65 |

# The Fix

[...]

PR #784, opened 2019-11-18
Corrected sign on `.resid`

CGMossa:
According to #778, the sign for residual calculation is flipped. I applied the change and made this PR.


alexpghayes:
I don't think the definition of the residual is incorrect. Please see comments in #778.

PR #790, opened 2019-11-23
BUG FIX: utils::globalVariables() needs to be imported

HenrikBengtsson:
Currently, reverse package dependency checks report:

```
Error in globalVariables(c(".", ".id", ".rownames", "ci.lower", "ci.upper",  :
  could not find function "globalVariables"
Error: unable to load R code in package ‘broom’
Execution halted
```

See https://github.com/HenrikBengtsson/future/blob/develop/revdep/problems.md for lots of examples of this.


alexpghayes:
Thank you!

PR #807, opened 2019-12-18
nest(data = -keep_col)

ChuliangXiao:
Argument `data` in function [nest()](https://tidyr.tidyverse.org/reference/nest.html) {tidyr} v1.0.0 should be explicit. Otherwise, there is a warning.

``` r
library(tidyr)
library(magrittr)
#> 
#> Attaching package: 'magrittr'
#> The following object is masked from 'package:tidyr':
#> 
#>     extract
nested <- Orange %>% 
  nest(-Tree)
#> Warning: All elements of `...` must be named.
#> Did you want `data = c(age, circumference)`?
```

<sup>Created on 2019-12-18 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

### DIFF
--- a/vignettes/broom_and_dplyr.Rmd
+++ b/vignettes/broom_and_dplyr.Rmd
@@ -71,7 +71,7 @@ library(tidyr)
 library(purrr)
 
 nested <- Orange %>% 
-  nest(-Tree)
+  nest(data = -Tree)
 ```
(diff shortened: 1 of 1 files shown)


simonpcouch:
Thanks for catching this!🙂

PR #833, opened 2020-04-16
Add option to return p.values to tidy.polr

LukasWallrich:
MASS::polr models by default do not provide p-values, as their t-values are not normally distributed. However, the MASS::dropmodel function offers a way to test the significance of terms in MASS::polr models (this is the prefered way of the MASS author to test for the significance of specific terms). This PR adds an option to tidy.polr to request such p.values.


alexpghayes:
Looks good. Can you update `NEWS` and add yourself as a contributor in `DESCRIPTION`?

LukasWallrich:
Done - thanks.

PR #840, opened 2020-04-24
GitHub Actions

topepo:
Hopefully this PR will kick-start them into working.


topepo:
There are some `broom` failures here but I want to check on a few things before we consider merging this.

alexpghayes:
Some of the broom failures are expected (in particular the augment.glm())
failure, which you can safely ignore.

topepo:
I fixed the one non-broom issue so this is good to go. 

FYI It runs the tests more often than the default (which is commit to master or PR to master). I find these settings more helpful and avoids a lot of draft PRs that are just to get the tests running.

FYI 2: for some reason, the unix build had

```
The following packages will be REMOVED:
  libomp-9-dev libomp5-9
The following NEW packages will be installed:
  libomp5
```

so I added two `apt-get` installs to the GHA workflow. Seems to work fine.

alexpghayes:
Is there a "Github Actions for the uninitiated R dev" guide somewhere that I can read to understand all this newfangled glory?

PR #842, opened 2020-05-07
Remove `Naive SE` in tidy.survreg

### LINKED ISSUE #728: tidy.survreg function fails when robust = T in survreg model
Error when using robust regression in survreg. I think the issue is that robust regression gives a summary table with an extra column (called "(Naive SE)"). 

```r
fmla <- as.formula(paste("Surv(time, status) ~ ","karno + age" ))
fit <- survreg(formula = fmla, data = veteran, dist = "lognormal", robust = T)
broom::tidy(fit)

summary(fit)$table
```

I propose changing the tidy.survreg function as follows to fix this error

```r
  tidy.survreg <- function(x, conf.level = .95, conf.int = FALSE, ...) {
  s <- summary(x)
  # add extra column name here
  nn <- c("estimate", "std.error", "(Naive SE)", "statistic", "p.value")
  ret <- fix_data_frame(s$table, newnames = nn)
  
  if(conf.int){
    # add confidence interval
    ci <- stats::confint(x, level = conf.level)
    colnames(ci) <- c("conf.low", "conf.high")
    ci <- fix_data_frame(ci)
    ret <- as_tibble(merge(ret, ci, all.x = TRUE, sort = FALSE))
  }
  
  ret
  }
```

karldw:
I added code to drop the `(Naive SE)` column when `survreg` is run with `robust=TRUE`. This extra column was causing the issues in #728. It would be possible to include it, but that seems less tidy.

I'm having trouble running  the full set of tests, but the survival tests seem fine.

Obviously there are bigger things going on right now. This PR isn't urgent.

Fixes #728

### DIFF
--- a/R/survival-survreg-tidiers.R
+++ b/R/survival-survreg-tidiers.R
@@ -4,13 +4,13 @@
 #' @param x An `survreg` object returned from [survival::survreg()].
 #' @template param_confint
 #' @template param_unused_dots
-#' 
+#'
 #' @evalRd return_tidy(regression = TRUE)
 #'
 #' @examples
 #'
 #' library(survival)
-#' 
+#'
 #' sr <- survreg(
 #'   Surv(futime, fustat) ~ ecog.ps + rx,
 #'   ovarian,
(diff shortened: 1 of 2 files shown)


simonpcouch:
Looks great to me! Thanks.🙂

PR #852, opened 2020-05-20
transition to new tidyr API

simonpcouch:
Some edits to transition away from `gather()` and `spread()` to `pivot_longer()` and `pivot_wider()`. This PR also removes the {reshape2} dependency!🐙

Will wait to merge until @topepo has a chance to run another round of revdepcheck on `master`.


simonpcouch:
Thank you for the pointers @jennybc! Will take a look and make adjustments--looks there are only a few uses of scoped dplyr verbs in the package right now.🙂

PR #877, opened 2020-06-10
Updates to fixest tidiers

karldw:
- Fix a bug in `fixest.augment` with new data: now correctly return predictions (previously returned residuals)
- Clarify allowed types of predictions and residuals
- The `residuals.fixest` method was introduced in fixest v0.5.0, so increase the required version

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -534,7 +534,7 @@ Suggests:
     emmeans,
     epiR,
     ergm,
-    fixest (>= 0.3.1),
+    fixest (>= 0.5.0),
     gam (>= 1.15),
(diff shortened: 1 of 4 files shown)


simonpcouch:
Now that `0.7.0rc` is merged back into `master`, could you merge upstream? We made some adjustments to these tidiers due to a release of the package that was happening at the same time as we were submitting, so some of these might have been addressed (if less thoroughly.)

Thanks!

karldw:
Done! Let me know if there are other changes you'd like to see.

simonpcouch:
Looks great, thanks! Merging.

Issue #829, opened 2020-04-07
Allow for changing the RMST produced in glance.survfit


asondhi:
glance.survfit (and presumably tidy.survfit) returns a dataframe of summary statistics for a survfit object, but does not allow for specifying the restricted mean survival time (RMST) threshold (which is handled by the print.survfit function in the survival package), and instead only can produce the RMST using default threshold.  

```r
library(tidyverse)
library(broom)
library(survival)

T = rexp(1000, rate = 1)
status = rbinom(1000, 1, 0.6)
sfit = survfit(Surv(time = T, event = status) ~ 1)

print(sfit, print.rmean = TRUE) # print summary stats with default threshold for RMST
sfit %>% glance # matches the above print 

print(sfit, rmean = 3) # different threshold for RMST
sfit %>% glance(rmean = 3) # still uses default threshold
```

vincentarelbundock:
This can be achieved by passing `glance`'s `...` argument to the `summary(x, ...)` function that we use to extract information from the `survfit` model, as illustrated in this PR:

https://github.com/tidymodels/broom/pull/880

simonpcouch:
Closed in #880!

PR #880, opened 2020-06-12
glance.survfit: allow RMST adjustment

vincentarelbundock:
Issue https://github.com/tidymodels/broom/issues/829 requests a change to `glance.survfit` to allow adjustment of the restricted mean survival time (RMST) threshold:

This can be achieved by passing `glance`'s `...` argument to the `summary(x, ...)` method that we use to retrieve information from the `survfit` object. 

This PR changes a single line of code to achieve the desired result:

``` r
library(tidyverse)
library(broom)
library(survival)

T = rexp(1000, rate = 1)
status = rbinom(1000, 1, 0.6)
sfit = survfit(Surv(time = T, event = status) ~ 1)

print(sfit, print.rmean = TRUE) # print summary stats with default threshold for RMST
#> Call: survfit(formula = Surv(time = T, event = status) ~ 1)
#> 
#>          n     events     *rmean *se(rmean)     median    0.95LCL    0.95UCL 
#>    1000.00     592.00       1.69       0.08       1.19       1.07       1.32 
#>     * restricted mean with upper limit =  8.47
sfit %>% glance # matches the above print 
#> # A tibble: 1 x 10
#>   records n.max n.start events rmean rmean.std.error median conf.low conf.high
#>     <dbl> <dbl>   <dbl>  <dbl> <dbl>           <dbl>  <dbl>    <dbl>     <dbl>
#> 1    1000  1000    1000    592  1.69          0.0800   1.19     1.07      1.32
#> # … with 1 more variable: nobs <int>

print(sfit, rmean = 3) # different threshold for RMST
#> Call: survfit(formula = Surv(time = T, event = status) ~ 1)
#> 
#>          n     events     *rmean *se(rmean)     median    0.95LCL    0.95UCL 
#>   1.00e+03   5.92e+02   1.40e+00   3.74e-02   1.19e+00   1.07e+00   1.32e+00 
#>     * restricted mean with upper limit =  3
sfit %>% glance(rmean = 3) # still uses default threshold
#> # A tibble: 1 x 10
#>   records n.max n.start events rmean rmean.std.error median conf.low conf.high
#>     <dbl> <dbl>   <dbl>  <dbl> <dbl>           <dbl>  <dbl>    <dbl>     <dbl>
#> 1    1000  1000    1000    592  1.40          0.0374   1.19     1.07      1.32
#> # … with 1 more variable: nobs <int>
```

<sup>Created on 2020-06-12 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>


vincentarelbundock:
I merged upstream master into this PR to show that it still works. This commit only includes 4 characters and closes this issue: https://github.com/tidymodels/broom/pull/880

simonpcouch:
Ah, thanks for the nudge. Could you also write custom documentation here to note that it's not actually ignored? The documentation is currently the default for `...` elsewhere in the package: 

```
Additional arguments. Not used. Needed to match generic signature only. {etc...}
```

vincentarelbundock:
Sure. I tried to model the new doc on another one I read, but let me know if you prefer a different format/content.

I force-pushed to keep the ocmmit clean.

simonpcouch:
Great, thanks!

PR #895, opened 2020-07-12
add lavaan:: before lavIspect

milanwiedemann:
This fixes an error message saying that the `lavInspect` function wasn't found (only when the `lavaan` package isn't loaded).


simonpcouch:
Much appreciated!

Issue #578, opened 2019-01-10
Variable names do not match in the contents: tidy.prcomp(matrix = "pcs")


GegznaV:
In the results of `tidy.prcomp` (with `matrix = "pcs"`), variable name `percent` is misleading. Documentaion states:

![image](https://user-images.githubusercontent.com/12725868/50937948-ae0c7e80-147e-11e9-9e2d-af6e30803937.png)

I understand "percent" as a number ranging from 0 to 100, and the result is the number ranging from 0 to 1 (i.e., a proporton or a fraction). In the example below, `0.620` is 62.0% not 0.62%. In some situations it is confusing.

``` r
library(broom)
pc <- prcomp(USArrests, scale = TRUE)
tidy(pc, "pcs")
#> # A tibble: 4 x 4
#>      PC std.dev percent cumulative
#>   <dbl>   <dbl>   <dbl>      <dbl>
#> 1     1   1.57   0.620       0.620
#> 2     2   0.995  0.247       0.868
#> 3     3   0.597  0.0891      0.957
#> 4     4   0.416  0.0434      1
```

<sup>Created on 2019-01-10 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>

What are possible solutions to remove the ambiguity? Rename from "percent" to "fraction". Multiply the result by 100? I think the multiplication would be a better solution as a printed result is less cluttered with leading zeros. Compare:

![image](https://user-images.githubusercontent.com/12725868/50938339-6c7cd300-1480-11e9-9cb0-bcfa2b7a2342.png)
vs.
![image](https://user-images.githubusercontent.com/12725868/50938355-7b638580-1480-11e9-8e03-cbd6703b6c44.png)

deblnia:
Hi, @alexpghayes and @GegznaV, I'd love to take this on!

GegznaV:
I still think, that word "percent" and value "1" are misleading, as as they imply "1%".
If itbis a

GegznaV:
Still, if no breaking changes are wanted to be intoduced, please, add note in documentation, that "1" means "100%".

alexpghayes:
@deblnia this is the issue I mentioned today

PR #901, opened 2020-07-19
Changed `tidy.prcomp()` to be whole numbers (#578) 

deblnia:
# Hi!

I've multiplied the `tidy.prcomp()` method by 100 so percentages are shown as whole numbers instead of decimals (as per #578 by GegznaV). The change is documented in `NEWS.md`, I've added myself as a contributor, and R-CMD doesn't seem to have added any additional warnings (though I might be wrong on this, happy to correct if so).

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -486,7 +486,11 @@ Authors@R:
       person(given = "Matthieu",
              family = "Stigler",
              role = "ctb",
-             email = "Matthieu.Stigler@gmail.com"))
+             email = "Matthieu.Stigler@gmail.com"),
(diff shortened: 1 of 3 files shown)


simonpcouch:
Hi there! Talked with Alex briefly about this earlier, and we felt that these changes might result in unnecessary breakages.

How would you feel about, instead, adjusting the documentation to clarify that the percentages are represented as decimals on [0, 1]?

deblnia:
Sounds good! Will revise PR 😄

simonpcouch:
Closed with #912.

PR #910, opened 2020-08-12
Fix tidy.prcomp labels for loadings and scores matrices

tavareshugo:
Hi, 

I think there was a bug in `tidy.prcomp()` related to [this commit](https://github.com/tidymodels/broom/commit/6e08bfd9a20ddd4d359cc55cc9bf830093d854dc#diff-8efa72cafc8cc137996521f2ae7b5c7a), which moved code from using `dplyr::gather()` to `dplyr::pivot_longer()`. These two functions sort the output tibble in different ways, and so the labels being added with `rep()` were wrong when using `tidy(x, matrix = "scores")` and `tidy(x, matrix = "loadings")`. 

Here's an example (using broom v0.7.0):

```r
library(magrittr)
set.seed(1)
m <- matrix(rnorm(8), ncol = 4)
rownames(m) <- paste0("row", 1:nrow(m))
colnames(m) <- paste0("col", 1:ncol(m))

x <- prcomp(m)

broom::tidy(x, "scores") %>% dplyr::arrange(PC)
```

Results in 

```
# A tibble: 4 x 3
  row      PC     value
  <chr> <dbl>     <dbl>
1 row1      1 -1.41e+ 0
2 row1      1  1.41e+ 0
3 row2      2  0.      
4 row2      2  2.08e-17
```

The result we want is rather: 

```
  row      PC     value
  <chr> <dbl>     <dbl>
1 row1      1 -1.41e+ 0
2 row2      1  1.41e+ 0
3 row1      2  0.      
4 row2      2  2.08e-17
```

I guess I could  have just changed `rep()` to use `each = ncomp` rather than `times = ncomp`. 
However, I think my fix is perhaps a bit simpler to read and more robust to future changes in how `pivot_longer()` sorts output. I took advantage of `tibble::as_tibble()` ability to add rownames as columns, and this works even if `rownames(x)` is `NULL`. 

I hope the PR is helpful. (And thanks for the amazing package!)


R/stats-prcomp-tidiers.R
```diff
@@ -108,38 +108,17 @@ tidy.prcomp <- function(x, matrix = "u", ...) {
...
-      as.data.frame()
-    ret <- data.frame(
-      label = rep(labels, times = ncomp),
-      variables,
-      stringsAsFactors = FALSE
-    )
-    names(ret) <- c("column", "PC", "value")
+    ret <- x$rotation %>% 
+      tibble::as_tibble(rownames = "labels") %>% 
+      tidyr::pivot_longer(cols = -"labels",
```
tavareshugo, line 113: apologies, I've misnamed the output column here, which I think triggered a failed test. I think this would be the correct name:

```suggestion
      tibble::as_tibble(rownames = "column") %>% 
      tidyr::pivot_longer(cols = -"column", 
```

R/stats-prcomp-tidiers.R
```diff
@@ -108,38 +108,17 @@ tidy.prcomp <- function(x, matrix = "u", ...) {
...
-      label = rep(labels, times = ncomp),
-      variables,
-      stringsAsFactors = FALSE
-    )
-    names(ret) <- c("column", "PC", "value")
+    ret <- x$rotation %>% 
+      tibble::as_tibble(rownames = "column") %>% 
+      tidyr::pivot_longer(cols = -"column", 
+                          names_to = "PC", 
+                          values_to = "value")
```
tavareshugo, line 115:
```suggestion
                          values_to = "value")
    if (is.null(rownames(x$rotation))) ret$column <- as.integer(ret$column)
```

R/stats-prcomp-tidiers.R
```diff
@@ -108,38 +108,17 @@ tidy.prcomp <- function(x, matrix = "u", ...) {
...
-    ret <- data.frame(
-      label = rep(labels, times = ncomp),
-      samples
-    )
-    names(ret) <- c("row", "PC", "value")
+    ret <- x$x %>% 
+      tibble::as_tibble(rownames = "row") %>% 
+      tidyr::pivot_longer(cols = -"row", 
+                          names_to = "PC", 
+                          values_to = "value")
```
tavareshugo, line 121:
```suggestion
                          values_to = "value")
    if (is.null(rownames(x$x))) ret$row <- as.integer(ret$row)
```

simonpcouch:
Thanks for pointing this out @tavareshugo! 

Should I consider this PR ready to review? How should I interpret your review comments/suggestions?

tavareshugo:
Hi @simonpcouch, thanks for having a look at this. I've committed my previous suggestions (i.e. for consistency with previous versions it returns labels as integer if rownames of those matrices are missing). I've also made a new commit adding my details to the contributors and a note to `NEWS.md`. So, I would say that this is ready for review now, thanks.

simonpcouch:
Thanks for your patience with this, @tavareshugo. This code looks great to me, so I'll go ahead and resolve conflicts and merge! Much appreciated.

tavareshugo:
thanks a lot @simonpcouch, I'm glad I could help a little bit with the package

Issue #921, opened 2020-08-31
Column name '.pred' for prediction intervals?


bwiernik:
The interval column names in augment() are always called `.conf.low` and `.conf.high`, regardless of whether `"confidence"` or `"prediction"` is specified for `interval`. I think this is potentially confusing and may be misleading in reports if the user doesn't rename the columns. I suggest changing the column names to `.pred.low` and `.pred.high` when `interval = "prediction"` (cf. in the metafor package, confidence intervals are labeled `ci` and prediction intervals labeled `pi`: https://github.com/wviechtb/metafor/blob/acacdcaa56ce6ed9c7c3239c15007b14d03012c8/man/predict.rma.Rd#L47).

``` r
broom::augment(lm(mpg ~ hp + disp, data = mtcars), interval = "prediction")
#> # A tibble: 32 x 12
#>    .rownames   mpg    hp  disp .fitted .conf.low .conf.high .resid .std.resid
#>    <chr>     <dbl> <dbl> <dbl>   <dbl>     <dbl>      <dbl>  <dbl>      <dbl>
#>  1 Mazda RX4  21     110  160     23.1     16.6        29.7 -2.15      -0.702
#>  2 Mazda RX~  21     110  160     23.1     16.6        29.7 -2.15      -0.702
#>  3 Datsun 7~  22.8    93  108     25.1     18.6        31.7 -2.35      -0.776
#>  4 Hornet 4~  21.4   110  258     20.2     13.5        26.8  1.23       0.408
#>  5 Hornet S~  18.7   175  360     15.5      8.82       22.1  3.24       1.08 
#>  6 Valiant    18.1   105  225     21.3     14.7        27.9 -3.20      -1.06 
#>  7 Duster 3~  14.3   245  360     13.7      7.02       20.4  0.575      0.194
#>  8 Merc 240D  24.4    62  147.    24.7     18.1        31.4 -0.344     -0.115
#>  9 Merc 230   22.8    95  141.    24.1     17.5        30.7 -1.30      -0.428
#> 10 Merc 280   19.2   123  168.    22.6     16.1        29.1 -3.39      -1.11 
#> # ... with 22 more rows, and 3 more variables: .hat <dbl>, .sigma <dbl>,
#> #   .cooksd <dbl>
```

<sup>Created on 2020-08-31 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

simonpcouch:
[...]

PR #925, opened 2020-09-04
Change names of interval columns in augment() for clarity

bwiernik:
Closes https://github.com/tidymodels/broom/issues/921

- Changes `.conf.low` to `.lower` and `.conf.high` to `.upper` consistently for `augment()` output. 
- Updates `agument.rma()` to consistently return one set of intervals.


R/rma-tidiers.R
```diff
@@ -235,9 +235,9 @@ augment.rma <- function(x, ...) {
   pred <- as.data.frame(pred)
 
   # fix names
-  names(pred)[1:4] <- c(".fitted", ".se.fit", ".conf.low", ".conf.high")
-  credible_intervals <- names(pred) %in% c("cr.lb", "cr.ub")
-  names(pred)[credible_intervals] <- c(".cred.low", ".cred.high")
+  names(pred)[1:4] <- c(".fitted", ".se.fit", ".conf.lower", ".conf.upper")
```
simonpcouch, line 238: Since these don't end up being fully internally consistent, I'd prefer we stick with the current column naming so as to avoid breaking changes.

simonpcouch:
Thanks for making this happen!

I'm feeling positive about how this came together with respect to the first bullet.

I dropped a comment in about the second.

Could you ensure that changes have been made only to relevant documentation? Quite a few `.Rd` files with changes only in the help-files they link out to.

R/rma-tidiers.R
bwiernik: The developer of metafor has acknowledged that "cred"/"credibility" was not an accurate label for the column (they are prediction intervals) and has changed its column labels from "ci" to "pi". These values are definitely not credibility intervals, so the `.cred.low` and `.cred.high` labels should at least change to `pred.low` and `.pred.high` to match what the values actually are and the metafor API.

bwiernik: After that change to remove the inaccurate and misleading label on the prediction intervals, my though was that using the same "lower"/"upper" terminology as the other augment methods made sense.

bwiernik:
When I try to run devtools::document(), it doesn't seem to be producing the correct output (e.g., the Value table disappears entirely for agument.rma; and the Value table for augment.lm is missing the interval arguments). I think this has to do with the `return_arguments()` helper function, but I'm not sure what I need to do to fix it.

R/rma-tidiers.R
simonpcouch: Would it be possible to label these just `.lower` and `.upper`? More comfortable with the potential breaking change if we're being internally consistent.

bwiernik: There are two sets of intervals returned from `predict.rma()`—confidence intervals ("ci.lb", "ci.ub") and prediction intervals ("pi.lb", "pi.ub"). Unlike `predict.lm()`, `predict.rma()` always returns both. How should we handle that?

simonpcouch:
That's fine.👍

The other modified files are a result of differences in package versions. Recent switch in how documentation gets linked from place to place that can be a pain in package dev.

R/rma-tidiers.R
simonpcouch: Ah, right! Yeah, this is tough. Not sure if this was where your hesitancy came from originally, @alexpghayes.

bwiernik: I see two possible options here:

1. The path moving forward could be: intervals are always just labeled `.lower` and `.upper` if there is only one set. They are prepended with a consistent prefix (`.conf` for confidence, `.pred` for prediction, `.cred` for credible) if and only if the augment function returns multiple intervals at once. That ensures both consistency and accuracy in labeling (vs. the status quo when the `.conf.low` label is inaccurate in augment.lm when `interval="prediction"`).
1. Only return one interval from augment.rma() rather than two. That would be easy enough and make the behavior more consistent with other augment/predict methods.

bwiernik: On further review, this is a problem less than I actually thought. It's only an issue with the `rma.mv` class in metafor, which is fairly rarely used. The most common class `rma.uni` always returns prediction intervals the way `augment.rma` works. `rma.mh` and `rma.peto` only compute confidence intervals. 

So, with that in mind, I suggest we use `.lower` and `.upper` consistently. Then, for `rma.mv` only return the prediction interval (this makes the behavior consistent with rma.uni and is generally most appropriate in meta-analysis). We could add an `interval` argument to choose confidence instead.

bwiernik: Okay, I've implemented that. Now, all augment methods return `.lower` and `.upper` consistently.

simonpcouch:
Great, much appreciated!

There are some `modeltests` errors (along the lines of `Output column names not in the column glossary`) and some code/doc mismatch failures (since I had you roll back docs) that are fine to ignore from your end. Looks like there's still one failure in the examples that just requires some column renaming in a ggplot call--could you make that happen?

I'm feeling positive about where this is at. Looping you in, @alexpghayes.

bwiernik:
I’ll take a look tomorrow afternoon.

bwiernik:
@simonpcouch The example errors are just because the documentation Rd is out of sync with the roxygen. In the roxygen, all of the ggplot calls refer to `.lower` and `.upper`, but because I didn't run document() because of the modeltestr mismatch, they are erroring.

bwiernik:
@simonpcouch Did you want me to change anything regarding those errors or is this fine for now?

simonpcouch:
Oh, gotcha! No worries on the errors, then. Thanks for checking in.

simonpcouch:
Thanks for your patience!

Alex and I finally had a chance to talk on this, and we're game. Thanks for the contribution. Will go ahead and fix conflicts and then merge.

bwiernik:
Thanks so much! Glad it got merged!

PR #927, opened 2020-09-08
ivreg

### LINKED ISSUE #922: ivreg tidiers
The new [**ivreg**](https://john-d-fox.github.io/ivreg/index.html) package spins off the old `AER::ivreg()` methods into their own dedicated library.

The [existing](https://github.com/tidymodels/broom/blob/master/R/aer-tidiers.R) `ivreg()` tidiers all continue to work with the new package. But these should be separated from the other AER tidiers into their own file... especially since **ivreg** will become a dependency for **AER** in the next release (see [here](https://twitter.com/AchimZeileis/status/1301454613200154624)).

grantmcdermott:
Fixes #922

Highlights:
- After some consideration, I've created separate **ivreg** and **AER** tidiers. The `AER::ivreg` classes should inherit from the new `ivreg::ivreg` ones anyway. But splitting things up also allowed/prompted me to add a tidy method for AER::tobit, which now works instead of just producing an error.
- I've added support for an interval argument in `ivreg.augment`. Following #925 (currently pending) I've gone with ".lower" and ".upper" for the column names, but we'll just wait to see how that shakes out. This will obviously trigger an error on the CI build until the **modeltests** glossary is updated.
- All these new **ivreg** tidiers are also able to support "vcov" arguments if passed through `...`. Personally, I think this is pretty cool, because (a) it allows users to adjust standard errors, etc. on the fly and (b) it makes it consistent with a major goal of the new **ivreg** package (i.e. integrated supported for libraries like **lmtest** and **sandwhich**). 

Some examples adapted from the help docs:

```r
library(broom)
library(ggplot2)
library(ivreg) ## >=0.5.1

m <- ivreg(log(packs) ~ log(rprice) + log(rincome) | salestax + log(rincome),
           data = CigaretteDemand)

tidy(m)
#> # A tibble: 3 x 7
#>   term         estimate std.error statistic      p.value
#>   <chr>           <dbl>     <dbl>     <dbl>        <dbl> 
#> 1 (Intercept)     9.43      1.36      6.94  0.0000000124
#> 2 log(rprice)    -1.14      0.359    -3.18  0.00266
#> 3 log(rincome)    0.215     0.269     0.799 0.429

# We can choose which component of the the 2SLS we want to tidy with the
# "component" argument. For example, to only tidy the instrument(s):
tidy(m, component = "instrument")
#> # A tibble: 1 x 5
#>   term     estimate std.error statistic      p.value
#>   <chr>       <dbl>     <dbl>     <dbl>        <dbl>
#> 1 salestax   0.0274   0.00408      6.72 0.0000000265

# Example of using "vcov" to specify an alternate error structure. Here we use 
# HC (i.e. robust) standard errors from the sandwich package.
tidy(m, vcov = sandwich::vcovHC)
#> # A tibble: 3 x 7
#>   term         estimate std.error statistic      p.value
#>   <chr>           <dbl>     <dbl>     <dbl>        <dbl>
#> 1 (Intercept)     9.43      1.35      6.97  0.0000000115
#> 2 log(rprice)    -1.14      0.403    -2.84  0.00682
#> 3 log(rincome)    0.215     0.332     0.646 0.521

# Example that combines "interval" and "vcov" arguments in augment. Useful for 
# visually inspecting the impact of different  standard errors on our model 
# predictions.
 augment(m, newdata = CigaretteDemand[1:10,], 
         interval = 'confidence', vcov = sandwich::vcovHC) %>%
   ggplot(aes(x=.rownames, y=.fitted, ymin=.lower, ymax=.upper)) + 
   geom_pointrange() +
   ylim(4, 5) +
   labs(title = "Predicted values; robust standard errors")
```

<sup>Created on 2020-09-07 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

Lastly, note that some of these tidiers rely on v. 0.5.1 of the parent ivreg package. As of the time of writing, this hasn't hit CRAN yet, though I expect it will fairly soon. I didn't want to add a "remotes" call to the [dev version](https://john-d-fox.github.io/ivreg/) in the Description files, but let me know. I'm fine to have this PR sitting around for a bit if the upstream CRAN release is paramount. (Although, note that the existing ivreg tidiers will break in the meantime as documented [here](https://github.com/tidymodels/broom/issues/922#issuecomment-687925576).)


simonpcouch:
Thanks for making this happen! Feeling really positive about how these came together.

We're starting to think about reverse dependency checks and such for 0.7.1, so I'll hold off on merging until `ivreg` 1.5.1 and the version of `AER` that takes on `ivreg` as a dependency make it up on CRAN.

vincentarelbundock:
This looks really great; thanks for all the efforts!

FWIW, one of the `ivreg` contributors recently wrote on twitter that it might take a while for AER to switch over. So it probably makes sense to include those new tidiers before then.

Also, since `component="instruments"` is already a convenience subseting-type of feature, I wonder if it would make sense to also allow both stages, or stage2+instruments, either with a new component name, or by setting `several.ok=TRUE` in `match.arg`.

Again, thanks for this. I look forward to using it!

Issue #692, opened 2019-04-11
Consistency of emmeans tidiers (with other tidiers)


crsh:
While working on a [PR to add a new `emmeans`-tidier](https://github.com/tidymodels/broom/pull/691), I noticed that the `emmeans`-tidiers have some internal and external inconsistencies:

1. The [`lsmobj`-method](https://github.com/tidymodels/broom/blob/57b2400e4df99976f677a1dc360b54793a1232aa/R/emmeans-tidiers.R#L71) uses the common arguments `conf.int` and `conf.level` as defined in the `param_confint` template. The other methods (e.g., [`emmGrid`](https://github.com/tidymodels/broom/blob/master/R/emmeans-tidiers.R#L125)) do not provide these arguments and instead rely on the argument names native to the `emmeans` `summary()`-methods (e.g., `infer` and `level`).

I haven't looked exhaustively at other methods, but I additionally noticed some inconsistencies compared to other contrast tidiers, specifically [`tidy.TukeyHSD()`](https://github.com/tidymodels/broom/blob/57b2400e4df99976f677a1dc360b54793a1232aa/R/stats-anova-tidiers.R#L265):

2. `tidy.TukeyHSD()` reports the contrasted conditions in a column labelled `comparison` in the form of `a-b`. In contrast, the `emmeans` tidiers return the same information in two columns labelled `level1` and `level2` (containing `a` and `b`).

~~~r
fm1 <- aov(breaks ~ wool + tension, data = warpbreaks)
thsd <- TukeyHSD(fm1, "tension", ordered = TRUE)
tidy(thsd)

emmp <- pairs(emmeans(fm1, ~ tension))
tidy(emmp)
~~~

3. In the `tibble` returned by `tidy.TukeyHSD()`, the column containing p-values is labelled `adj.p.value`. In contrast, the `emmeans` tidiers label this column `p.value` regardless of whether it has been adjusted for multiple comparisons or not (see code above). Unless I missed something, the use of `adj.p.value` is currently unique to `tidy.TukeyHSD()`. 

It seems desirable to try to keep things consistent across methods where possible but particularly within the set of tidiers for a given package. I would, therefore, suggest the following changes, that I'd be willing to implement in a PR:

1. Add the arguments `conf.int` and `conf.level` to all emmeans tidiers.
2. Change reporting of contrast pairs in either `tidy.TukeyHSD()` or the `emmeans`-methods. I'm not sure which of the two is preferable here.
[...]

PR #941, opened 2020-09-25
Map `effect.size` to `estimate` in emmeans tidiers

crsh:
[As of version 1.4.2](https://github.com/rvlenth/emmeans/blob/master/NEWS.md#emmeans-142) `emmeans` provides a new function `eff_size()` that returns an `emmGrid`-object where the column that contains the estimate is named `effect.size`. Because the current tidier has no mapping for this column it is not appropriately renamed in the tiedied output.

~~~r
library("broom")
library("emmeans")
 
fiber.lm <- lm(strength ~ diameter + machine, data = fiber)
emm <- emmeans(fiber.lm, "machine")
es <- eff_size(emm, sigma = sigma(fiber.lm), edf = df.residual(fiber.lm))
es
~~~
~~~
 contrast effect.size    SE df lower.CL upper.CL
 A - B         -0.650 0.650 11   -2.081    0.781
 A - C          0.993 0.726 11   -0.604    2.590
 B - C          1.643 0.800 11   -0.118    3.405

sigma used for effect sizes: 1.595 
Confidence level used: 0.95 
~~~
~~~r
tidy(es)
~~~
~~~
# A tibble: 3 x 7
  level1 level2 effect.size std.error    df conf.low conf.high
  <chr>  <chr>        <dbl>     <dbl> <dbl>    <dbl>     <dbl>
1 A      B           -0.650     0.650    11   -2.08      0.781
2 A      C            0.993     0.726    11   -0.604     2.59 
3 B      C            1.64      0.800    11   -0.118     3.40 
~~~

This pull request adds the missing mapping such that the `effect.size` column in the tidied output is renamed to `estimate`.

### DIFF
--- a/R/emmeans-tidiers.R
+++ b/R/emmeans-tidiers.R
@@ -183,6 +183,7 @@ tidy_emmeans_summary <- function(x, null.value = NULL, term_names = NULL) {
     "emmean" = "estimate",
     "pmmean" = "estimate",
     "prediction" = "estimate",
+    "effect.size" = "estimate",
     "SE" = "std.error",
     "lower.CL" = "conf.low",
     "upper.CL" = "conf.high",


IndrajeetPatil:
Wouldn't it also be better if the column `contrast` here is renamed to `term` (instead of it splitting into two separate columns), in order to be consistent with the other tidiers?

crsh:
See https://github.com/tidymodels/broom/issues/692 for a related discussion.

simonpcouch:
Much appreciated!

Issue #708, opened 2019-05-28
neither `tidy` nor `glance` methods work for `gee` objects


IndrajeetPatil:
Objects of class `gee` are not supported but because it also inherits `glm` class, I guess that's the tidier that gets picked up.

``` r
# setup
set.seed(123)
library(MASS)
library(gee)
data(OME)

# model
fm <-
  gee(
    cbind(Correct, Trials - Correct) ~ Loud + Age + OME,
    id = ID,
    data = OME,
    family = binomial,
    corstr = "exchangeable"
  )
#> Beginning Cgee S-function, @(#) geeformula.q 4.13 98/01/27
#> running glm to get initial regression estimate
#> (Intercept)        Loud         Age     OMEhigh      OMElow 
#> -5.90984676  0.15516993  0.01876291 -0.07569691 -0.31660542

# class of object
class(fm)
#> [1] "gee" "glm"

# summary
summary(fm)
#> 
#>  GEE:  GENERALIZED LINEAR MODELS FOR DEPENDENT DATA
#>  gee S-function, version 4.13 modified 98/01/27 (1998) 
#> 
#> Model:
#>  Link:                      Logit 
#>  Variance to Mean Relation: Binomial 
#>  Correlation Structure:     Exchangeable 
#> 
#> Call:
#> gee(formula = cbind(Correct, Trials - Correct) ~ Loud + Age + 
#>     OME, id = ID, data = OME, family = binomial, corstr = "exchangeable")
#> 
#> Summary of Residuals:
#>        Min         1Q     Median         3Q        Max 
#> -0.9585811  0.5507230  2.0414189  3.1210663 12.0241951 
#> 
#> 
#> Coefficients:
#>                Estimate  Naive S.E.     Naive z Robust S.E.    Robust z
#> (Intercept) -5.90100336 0.336455218 -17.5387482 0.231463368 -25.4943295
#> Loud         0.15507565 0.007513314  20.6401142 0.005321880  29.1392592
#> Age          0.01851319 0.003561640   5.1979385 0.003305525   5.6006793
#> OMEhigh     -0.04183516 0.160288285  -0.2609995 0.152182352  -0.2749015
#> OMElow      -0.28563252 0.131496124  -2.1721745 0.117510247  -2.4307031
#> 
#> Estimated Scale Parameter:  1.250466
#> Number of Iterations:  4
#> 
#> Working Correlation
#>              [,1]        [,2]        [,3]        [,4]        [,5]
[...]

PR #957, opened 2020-12-03
allow glm subclasses to use the glm tidiers

simonpcouch:
This PR adds in a similar warning for the `glm` tidiers as is implemented for `lm` tidiers: if `glm` is subclassed by some other class and dispatch sends to the `glm` tidiers, warn before attempting to run.

This PR also gets rid of the `gee <- tidy.default` assignment. The `glm` tidiers error out anyway, but perhaps the newly supplied warning

> Tidiers for objects of class gee are not maintained by the broom team,  and are only supported through the tidy.glm() method. Please be cautious in interpreting and reporting broom output. 

is more informative than the current `No tidy method for objects of class gee`. If this approach is go, we should also do away with the `tidy.default` assignment with other tidiers that subclass `lm` or `glm`.

Related to tidymodels/multilevelmod#5


simonpcouch:
I think, as implemented, this will warn on `glm`. Will come back to this in a sec.

asshah4:
As the `gee` tidiers will no longer be overwritten by the `tidy.default` assignment, a new tidier can now be made? With the `gee` function, it creates both classes (`gee` and `glm`) in an object. Without specifying directly with `tidy.gee()`, how do we specify the correct method to dispatch?

``` r
library(gee)
#> Warning: package 'gee' was built under R version 4.0.3
gee_mod <- gee(mpg ~ hp + wt, id = cyl, data = mtcars)
#> Beginning Cgee S-function, @(#) geeformula.q 4.13 98/01/27
#> running glm to get initial regression estimate
#> (Intercept)          hp          wt 
#> 37.22727012 -0.03177295 -3.87783074
class(gee_mod)
#> [1] "gee" "glm"
```

<sup>Created on 2020-12-03 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

simonpcouch:
> a new tidier can now be made?

Alex and I likely won't put in the time to write new tidiers here, but would be glad to review a PR if others would like to [make that happen](https://www.tidymodels.org/learn/develop/broom/). :-)

> how do we specify the correct method to dispatch?

This generally happens automatically. We've opted to adjust that behavior in the past, and are considering our options here.

Our current approach is to explicitly tell objects that subclass `glm` error out. The approach proposed in this PR would allow giving the `*.glm()` tidiers a go after raising a warning noting that output should be taken with a grain of salt. As of now, the `gee` tidiers would error out in that case, as `gee` objects don't have their own dedicated tidiers and don't behave like `glm` objects like broom needs them to.

Related to #708.

asshah4:
> Alex and I likely won't put in the time to write new tidiers here, but would be glad to review a PR if others would like to [make that happen](https://www.tidymodels.org/learn/develop/broom/). :-)

Sounds good! I'll put in a pull request in the near future. I'd like to be up and running with `gee` models in the near future. I'll see if I can make sure it dispatches appropriately now that it has become an option.

asshah4:
Added a new set of tidiers here for #958. They seem to be working locally. Thanks!

simonpcouch:
@alexpghayes Could you take a look at this ahead of the coming release? Essentially, proposes making the error -> warning switch for `glm` subclasses the same way we did for `lm`.

alexpghayes:
Yeah will do, probably on Friday.

R/stats-glm-tidiers.R
```diff
@@ -132,7 +135,8 @@ augment.glm <- function(x,
 #' @family lm tidiers
 #' @seealso [stats::glm()]
 glance.glm <- function(x, ...) {
-  check_for_subclasses(x)
+  check_glm_class(x)
```
alexpghayes, line 138: Can you change the name of this to something more informative? Maybe `check_for_broom_mixed_support()`?

alexpghayes (changes requested):

alexpghayes:
Other than the renaming comment I like this. One thing to note is that I used the `tidy.gee <- tidy.default` trick for several other classes. Eventually we should probably remove those in light of the new informative error approach. I will open an issue to remind us to do this.

simonpcouch:
Sure thing! I'll work through #959 when I address your renaming comment.

PR #958, opened 2020-12-07
Adding `gee` tidiers

asshah4:
Added tidy, glance, and basic augment function for gee. However, as `predict()` does not currently work for `gee::gee()`, nor does the model object return the predictor values, augment is limited. In addition, `glance()` can be improved upon if there are some agreement with how to assess quality of these clusters (such as a Hosmer-Lemeshow test).

Please let me know if anything else is needed.


asshah4:
Hi, @simonpcouch I'm struggling to make sure this will merge. I haven't installed all the dependent libraries for the tests (each individual modeling package), which I figured shouldn't cause an error during a CRAN check (but may error locally). However, there is a package dependency called `lfe` that shows up. I'm not sure how to handle this. 

My commit to the `gee-tidiers.R` is the only change I've made, so not sure if you've seen this error popup. 

```r
> checking package dependencies ... ERROR
  Package suggested but not available: 'lfe'
  
  The suggested packages are required for a complete check.
  Checking can be attempted without them by setting the environment
  variable _R_CHECK_FORCE_SUGGESTS_ to a false value.
  
  See section 'The DESCRIPTION file' in the 'Writing R Extensions'
  manual.

1 error x | 0 warnings v | 0 notes v
Error: Error: R CMD check found ERRORs
Execution halted
Error: Process completed with exit code 1.
```

simonpcouch:
Thanks for the PR! Well done.

You're fine to ignore the `lfe` error—that package was (surprisingly) archived by CRAN this morning. I'm not seeing any new errors/warnings/notes that seem to be from your end.🌼

I'm seeing 22 files changed at the moment, mostly for documentation updates for packages other than `gee`. Those are likely due to differing installed package versions. Could you roll back `.Rd` changes that aren't related to `gee`?

I'm not able to review the PR more fully at the moment, but will get to this as soon as I can!

simonpcouch:
Also, could you give a go at unit tests for these tidiers using the `modeltests` package? An example of what this looks like in broom can be found [here](https://github.com/tidymodels/broom/blob/master/tests/testthat/test-stats-lm.R). :-)

simonpcouch:
Related to bbolker/broom.mixed#110—I'm wondering if these are better suited to broom.mixed.

asshah4:
Yeah, I'll add the unit tests as well - thanks for reminding me. 

About #bbolker/broom.mixed#110, it should likely go wherever `tidy.geeglm()` ends up going as well.

asshah4:
I believe this is working now - added tests as well. The area of weakness here I believe is `glance()` and `augment()`. I'll keep thinking about how to best show model fit summaries on GEEs, and see about writing a workaround for the lack of a `predict` dispatch method.

R/gee-tidiers.R
```diff
@@ -0,0 +1,135 @@
+#' @templateVar class gee
+#' @template title_desc_tidy
+#'
+#' @param x A `gee` object returned from a call to [gee::gee()].
+#' @param conf.int 	Logical indicating whether or not to include a confidence
```
alexpghayes, line 5: Would you use templates for parameter documentation for the standard arguments? See https://github.com/tidymodels/broom/blob/master/R/biglm-tidiers.R#L6 for an example

R/gee-tidiers.R
```diff
@@ -0,0 +1,135 @@
...
+#' @aliases gee_tidiers 
+#' @seealso [tidy()], [gee::gee()]
+tidy.gee <- function(x, conf.int = FALSE, conf.level = .95, exponentiate = FALSE, ...) {
+  
+  coefs <-
+    stats::coef(summary(x)) %>%
+    tibble::as_tibble(rownames = "term") %>%
+    dplyr::select(-c(`Naive S.E.`, `Naive z`)) 
+  
+  # Remove naive estimators
```
alexpghayes, line 56: Can you expand this comment to include why we should remove the naive estimators?

R/gee-tidiers.R
```diff
@@ -0,0 +1,135 @@
...
+  names(coefs) <- c("term", "estimate", "std.error", "statistic")
+  
+  if (conf.int) {
+    zstat <- qnorm((1 + conf.level)/2)
+    coefs$conf.low <- coefs$estimate - zstat * coefs$std.error
+    coefs$conf.high <- coefs$estimate + zstat * coefs$std.error
+  }
+  
+  if (exponentiate) {
```
alexpghayes, line 65: We have a new approach to coefficient exponentiation which it would be nice to use here. See https://github.com/tidymodels/broom/blob/master/R/biglm-tidiers.R#L57 for an example.

tests/testthat/test-gee-tidiers.R
```diff
@@ -0,0 +1,28 @@
+context("gee-gee")
```
alexpghayes, line 1: As of testthat version 3 calls to `context()` are no longer necessary :)

tests/testthat/test-gee-tidiers.R
```diff
@@ -0,0 +1,28 @@
...
+library(modeltests)
+
+test_that("gee tidier arguments", {
+  check_arguments(tidy.gee)
+  check_arguments(glance.gee)
+})
+
+g1 <- gee::gee(mpg ~ wt + hp, id = cyl, data = mtcars, corstr = "exchangeable")
+
+# Tidy
```
alexpghayes, line 13: I think this comment (and similarly for `glance`) is probably unnecessary as it is already very clear what your tests do :)

alexpghayes (changes requested):

Issue #945, opened 2020-10-07
Allow tidy_emmeans_summary() to print std.error and conf.low/high


jmbarbone:
`tidy.emmGrid() ` only allows for either a std.error or conf.int to returned.
Unsure if there is any reason but it looks like the code removes it but using `as.data.frame()` on the `emmGrid` object returns standard error and confidence intervals.

https://github.com/tidymodels/broom/blob/b7fed0d62c49891e63bbe4a1a373a044a5a84ca2/R/emmeans-tidiers.R#L211-L214

```r
library(broom) 
library(emmeans)

# linear model for sales of oranges per day
oranges_lm1 <- lm(sales1 ~ price1 + price2 + day + store, data = oranges)

# marginal averages
marginal <- emmeans(oranges_rg1, "day")
tidy(marginal)
# A tibble: 6 x 6
  day   estimate std.error    df statistic      p.value
  <chr>    <dbl>     <dbl> <dbl>     <dbl>        <dbl>
1 1         5.56      1.77    23      3.15 0.00451     
2 2         6.49      1.73    23      3.76 0.00103     
3 3        13.7       1.75    23      7.80 0.0000000658
4 4         8.74      1.73    23      5.04 0.0000420   
5 5        15.4       1.79    23      8.65 0.0000000110
6 6        11.4       1.77    23      6.45 0.00000140  

tidy(marginal, conf.int = TRUE)
# A tibble: 6 x 7
  day   estimate    df conf.low conf.high statistic      p.value
  <chr>    <dbl> <dbl>    <dbl>     <dbl>     <dbl>        <dbl>
1 1         5.56    23     1.91      9.22      3.15 0.00451     
2 2         6.49    23     2.92     10.1       3.76 0.00103     
3 3        13.7     23    10.0      17.3       7.80 0.0000000658
4 4         8.74    23     5.16     12.3       5.04 0.0000420   
5 5        15.4     23    11.7      19.1       8.65 0.0000000110
6 6        11.4     23     7.74     15.0       6.45 0.00000140  
```

Would prefer to be able to return both:
```r
# Just coercing the emmGrid returns SE and CI
as.data.frame(marginal)
  day    emmean       SE df  lower.CL  upper.CL
1   1  5.564415 1.768083 23  1.906856  9.221974
2   2  6.494807 1.728959 23  2.918183 10.071430
3   3 13.664571 1.751505 23 10.041308 17.287835
4   4  8.742289 1.733920 23  5.155403 12.329175
[...]

PR #962, opened 2020-12-20
emmeans tidier can return both std.error and conf.low at the same time

vincentarelbundock:
`tidy.emmGrid` drops the "std.error" column when `conf.int=TRUE`. This was reported here:

https://github.com/tidymodels/broom/issues/945

This PR removes the 3 lines of code that dropped the `std.error` column (and did nothing else), and adjusts the number of expected columns in tests.

### DIFF
--- a/R/emmeans-tidiers.R
+++ b/R/emmeans-tidiers.R
@@ -208,11 +208,6 @@ tidy_emmeans_summary <- function(x, null.value = NULL, term_names = NULL) {
 
   colnames(ret) <- dplyr::recode(colnames(ret), !!!(repl))
 
-  # Remove std.error if conf.low/high exist
-  if (any(c("conf.low", "conf.high") %in% colnames(ret))) {
-    ret <- select(ret, -std.error)
-  }
-
   # If contrast column exists, add null.value column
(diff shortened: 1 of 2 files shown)


simonpcouch:
Thanks for the PR!

I don't know much about this model type. @crsh, it looks like the relevant code comes from [a commit](https://github.com/tidymodels/broom/commit/d0f65b6c8e1ee79c561e48a4ca28e96573a0c39b) of yours—any arguments for why this ought to be kept around? Much appreciated!

crsh:
@simonpcouch I made all my changes with consistency across tidiers in mind (see https://github.com/tidymodels/broom/issues/692). I reckon, I did it this way, because the other tidiers either return standard errors or confidence intervals. But I understand this may not be desirable and other considerations may trump consistency here.

vincentarelbundock:
Interesting! I wonder which other tidiers only report one or the other. My experience is that, in the vast majority of cases, `conf.int=TRUE` will return both.

simonpcouch:
Ah, gotcha! The quick response is much appreciated, @crsh.

I'm on board for this PR. Thanks to both of you. :-)

Issue #568, opened 2019-01-03
confidence intervals for `garch` models are missing


IndrajeetPatil:
Confidence intervals are available for `garch` models, but they are missing from its respective `broom` tidier:
``` r
# setup
set.seed(123)
library(tseries)
#> Registered S3 method overwritten by 'xts':
#>   method     from
#>   as.zoo.xts zoo
#> Registered S3 method overwritten by 'quantmod':
#>   method            from
#>   as.zoo.data.frame zoo
data(EuStockMarkets)
dax <- diff(log(EuStockMarkets))[,"DAX"]
dax.garch <- tseries::garch(dax, trace = FALSE)

# summary
summary(dax.garch)
#> 
#> Call:
#> tseries::garch(x = dax, trace = FALSE)
#> 
#> Model:
#> GARCH(1,1)
#> 
#> Residuals:
#>       Min        1Q    Median        3Q       Max 
#> -12.18398  -0.47968   0.04949   0.65746   4.48048 
#> 
#> Coefficient(s):
#>     Estimate  Std. Error  t value Pr(>|t|)    
#> a0 4.639e-06   7.560e-07    6.137 8.42e-10 ***
#> a1 6.833e-02   1.125e-02    6.073 1.25e-09 ***
#> b1 8.891e-01   1.652e-02   53.817  < 2e-16 ***
#> ---
#> Signif. codes:  0 '***' 0.001 '**' 0.01 '*' 0.05 '.' 0.1 ' ' 1
#> 
#> Diagnostic Tests:
#>  Jarque Bera Test
#> 
#> data:  Residuals
#> X-squared = 12947, df = 2, p-value < 2.2e-16
#> 
#> 
#>  Box-Ljung test
#> 
#> data:  Squared.Residuals
#> X-squared = 0.13566, df = 1, p-value = 0.7126

# confidence intervals
confint(dax.garch)
#>           2.5 %       97.5 %
#> a0 3.157588e-06 6.120989e-06
#> a1 4.627785e-02 9.037965e-02
#> b1 8.566877e-01 9.214456e-01

# tidier output (no confidence intervals present)
broom::tidy(x = dax.garch, conf.int = TRUE)
#> # A tibble: 3 x 5
#>   term    estimate   std.error statistic  p.value
#>   <chr>      <dbl>       <dbl>     <dbl>    <dbl>
#> 1 a0    0.00000464 0.000000756      6.14 8.42e-10
[...]

PR #964, opened 2020-12-21
tidy.garch supports confint

vincentarelbundock:
`conf.int` and `conf.level` were missing before, as reported here:

https://github.com/tidymodels/broom/issues/568


simonpcouch:
Bravo. Thank you. :-)

Issue #771, opened 2019-10-19
augment error when some weights are equal to zero


vnijs:
I thought https://github.com/tidymodels/broom/pull/436 addressed errors when some weights might be equal to zero in `lm` or `glm` but that doesn't seem to be the case (see below). I also tried the latest version of the master branch on github but that gives another message (see below).

**Broom 0.5.2 on CRAN:**

```r
> wts <- c(0, rep(1, nrow(mtcars) - 1))
> mod <- lm(mpg ~ cyl + vs, weights = wts, data = mtcars)
> broom::augment(mod)
Error in `$<-.data.frame`(`*tmp*`, ".hat", value = c(`Mazda RX4 Wag` = 0.132696390658174,  : 
  replacement has 31 rows, data has 32
> 
> mod <- lm(mpg ~ cyl + vs, weights = wts, data = mtcars)
> broom::augment(mod)
Error in `$<-.data.frame`(`*tmp*`, ".hat", value = c(`Mazda RX4 Wag` = 0.132696390658174,  : 
  replacement has 31 rows, data has 32
```

**Master branch on github:**

```r
> wts <- c(0, rep(1, nrow(mtcars) - 1))
> mod <- lm(mpg ~ cyl + vs, weights = wts, data = mtcars)
> broom::augment(mod)
Error: Can't augment data with observation level measures.
Did you provide `data` with the exact data used for model fitting?
> 
> mod <- lm(mpg ~ cyl + vs, weights = wts, data = mtcars)
> broom::augment(mod)
Error: Can't augment data with observation level measures.
Did you provide `data` with the exact data used for model fitting?
```

simonpcouch:
Addressed in #965!

PR #965, opened 2020-12-21
augment.lm influence with zero weights

vincentarelbundock:
This issue reports that `augment.lm` breaks when some regression weights are equal to zero:

https://github.com/tidymodels/broom/issues/771

This PR fixes that and adds a new test.

The `add_hat_sigma_cols` in `R/utilities.R` was already designed to handle this corner case. I just moved `.std.resid` and `.cooksd` to that function. Cook's is zero when weight is zero because it is a measure of influence. The standardized residuals is set to `NA`.


simonpcouch:
Snazzy, thanks!🕺

Issue #867, opened 2020-06-02
Use exponentiate idiom when possible


simonpcouch:
Some tidiers methods don't make use of the exponentiate helper. 

```
  if (exponentiate) {
    ret <- exponentiate(ret)
  }
```

louisahsmith:
I came to report an inconsistency related to this, so will just post here, as using that helper would fix the issue. Tidying a `geeglm` object with `exponentiate = TRUE` doesn't exponentiate the CI, but does for `glm`:
``` r
library(broom)
library(geepack)

glm_mod <- glm(vs ~ mpg, data = mtcars, family = binomial())
gee_mod <- geeglm(vs ~ mpg, data = mtcars, family = binomial(),
                  corstr = "ind", id = 1:nrow(mtcars))
tidy(glm_mod, conf.int = TRUE, exponentiate = TRUE)
#> # A tibble: 2 x 7
#>   term        estimate std.error statistic p.value     conf.low conf.high
#>   <chr>          <dbl>     <dbl>     <dbl>   <dbl>        <dbl>     <dbl>
#> 1 (Intercept) 0.000146     3.16      -2.79 0.00522 0.0000000539    0.0202
#> 2 mpg         1.54         0.158      2.72 0.00659 1.20            2.28
tidy(gee_mod, conf.int = TRUE, exponentiate = TRUE)
#> # A tibble: 2 x 7
#>   term        estimate std.error statistic p.value conf.low conf.high
#>   <chr>          <dbl>     <dbl>     <dbl>   <dbl>    <dbl>     <dbl>
#> 1 (Intercept) 0.000146     3.11       8.06 0.00454  -14.9      -2.73 
#> 2 mpg         1.54         0.165      6.82 0.00899    0.107     0.753
```

<sup>Created on 2020-08-10 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

alexpghayes:
Just took a look and the `geepack` tidiers need a big refactor to move away from the `process_*()` idiom.

ddsjoberg:
I came to report this as well.  I'll add that the confidence interval not being exponentiated is a bug introduced with 0.7.0 (if that helps).  Thanks for all your work with broom!

``` r
packageVersion("broom")
#> [1] '0.5.6'
geeglm <- 
  geepack::geeglm(am ~ mpg, id = cyl, data = mtcars, family = binomial, corstr = "exchangeable")

[...]

PR #967, opened 2020-12-21
use exponentiate idiom where possible

vincentarelbundock:
Code consistency in the use of `exponentiate`. This PR addresses issue https://github.com/tidymodels/broom/issues/867

I grepped through all the tidiers to find non-standard uses of the `exponentiate` argument and I fixed all of them, except for `coxph`, which is dealt with in PR https://github.com/tidymodels/broom/pull/966


simonpcouch:
Well done, thanks. :-)

Issue #187, opened 2017-01-19
tidying coxph.penal class


KZARCA:
```r
library(survival)
pen <- coxph(Surv(time, status) ~ age + pspline(nodes), data = colon)
tidy(pen) #error
```
The problem is that `summary(pen)` dispatches to summary.coxph.penal.

simonpcouch:
Addressed in #969. :-)

PR #969, opened 2020-12-21
tidy.coxph and pspline

vincentarelbundock:
Issue https://github.com/tidymodels/broom/issues/187 reports that `tidy.coxph` fails when using a penalty through `pspline`.

This PR fixes this bug by modifying two lines of code and adding a test.


simonpcouch:
Bravo!

Issue #393, opened 2018-06-18
Laundry list of tidy argument / behavioral consistency items


alexpghayes:
Based on looking at the non-standard arguments to `tidy` methods.

- [x] `tidy.boot`: document that `conf.method` is passed to `boot::boot.ci(..., type = conf.method)`
- [ ] `tidy.brmsfit`. Rename `intervals = conf.int`, `prob = conf.level`, decide what to do with `parameters`
- [x] `tidy.felm` example returns NaN rather than NA. Remove `fe.error`, only use `fe`. Possibly rename to `include_fe`
- [ ] Document that undefined / unknown values should **always** be `NA`. Test for `Inf` and `NaN`.
- [ ] New tidier vignette: written documentation should default argument values
- [ ] New tidier vignette: document behavior when `data`/`newdata`/neither is provided
- [ ] `tidy.gam`: `parametric` currently a switch. should parametric always be returned, with non-parametric return optionally? Or should they both always be returned (leaning toward this). Ask Gavin.
- [ ] `tidy.kmeans`: rename `col.names` arg
- [x] `tidy.manova` doc: specify what `test` gets passed to. perhaps rename `test` to `method`, which is used elsewhere
- [ ] `tidy.prcomp` doc: rename `matrix` arg?
- [x] harmonize `diag` and `diagonal` across `tidy.dist` and `tidy.rcorr`
- [ ] `tidy.rjags`: `pars` consistency with other parameter type arguments. rename `estimate.method`?
- [ ] `tidy.rq/tidy.rqs`: rename `se.type` since everywhere else it's `std.error`. Remove `alpha` arg and use `conf.level` instead, inverting or whatever is necessary to translate between scales
- [ ] `tidy.stanfit`: remove `droppars`, redundant with `pars`. rename `rhat, ess` to `include_rhat, include_ess`? or just include these by default. I'm a fan of `tidy` giving you lots of stuff that you can then `select(-stuff)` away immediately after if you don't want it.
- [ ] `tidy.stanreg`: Decide on `parameters`, see below. Rename `intervals = conf.int`, `prob = conf.level`. #291.
- [x] `tidy.TukeyHSD`: remove `separate.levels` argument and allow user to call `dplyr::separate` later if they want. This is consistent with the `multcomp::glht` tidier behavior.

## Parameter selection in bayesian and hierarchical models

Currently arguments `pars`, `parameters`, `droppars` accept:

- a categorical choice defining which type of tidier to return
- a regex for a parameter names
[...]

PR #970, opened 2020-12-22
issue393 consistency issues

vincentarelbundock:
Fixes some minor code and argument consistency issues listed here:

https://github.com/tidymodels/broom/issues/393

### DIFF
--- a/R/btergm-tidiers.R
+++ b/R/btergm-tidiers.R
@@ -8,6 +8,7 @@
 #' @param x A [btergm::btergm()] object.
 #' @param conf.level Confidence level for confidence intervals. Defaults to
 #'   0.95.
+#' @template param_confint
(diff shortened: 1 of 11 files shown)


simonpcouch:
Thanks for the high quality work, as always!

While I agree that it'd be nice if the confidence interval arguments and their order were all internally consistent, I'm worried about unnecessary breaking changes here. Changes to `conf.int` defaults and argument ordering really came back to bite us in the 0.7.0 release process, and I'm not sure we have the maintenance capacity to make those kinds of changes again.

Will stew on this for a bit. :-)

vincentarelbundock:
No worries at all. Just trying to be helpful and I have no emotional attachment to this particular PR. Feel free to scrap it.

Issue #552, opened 2018-12-12
add tidier for competing risks results `crr` class from `cmprsk` package


margarethannum:
---

My colleagues and I use competing risks models often and the results seem to be ripe for a tidier! 

I have started working on the problem, using the survival-coxph-tidiers as a template since the results are quite similar. 

I would be happy to work on this issue - are there any other competing risks models that people commonly like to use?

 
``` r
library(cmprsk)
#> Loading required package: survival
set.seed(2)
#time to loco-regional failure (lrf)
lrf_time <- rexp(100)

# Event 0 = censored, 1 = LRF w/ or w/o death, 2 = Death w/o LRF
lrf_event <- sample(0:2, 100, replace = TRUE) 

# Treatment vs placebo or something
trt <- sample(0:1, 100, replace = TRUE)

# Another variable
strt <- sample(1:2, 100, replace = TRUE)

#make competing risks model 
mod_crr <- crr(lrf_time, lrf_event, cbind(trt, strt))

# very tidy-able summary of coefficients
cmprsk::summary.crr(mod_crr)$coef
#>             coef exp(coef)  se(coef)          z p-value
#> trt   0.11481318 1.1216639 0.3793701  0.3026416    0.76
#> strt -0.04375598 0.9571875 0.3876434 -0.1128769    0.91
```

<sup>Created on 2018-12-12 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>

simonpcouch:
Closing in favor of #971.🐣

PR #971, opened 2020-12-22
new tidier: cmprsk

vincentarelbundock:
Tidier requested here: https://github.com/tidymodels/broom/issues/552

The `glance` method is pretty bare, but I don't see a lot of other stuff stored in the object, and the standard methods like `AIC` and such don't work for this model class.

Summary reports this:

```r
Pseudo likelihood ratio test = 0.07  on 2 df,
```

But I'm not sure how to call that statistic in `glance`


simonpcouch:
Thanks!!

I'm not familiar with this model type, and am also not sure how to make sense of / situate that last line in the print summary. I poked around the source code and documentation briefly but couldn't glean much.

@margarethannum, I know it's been quite a while since you filed your issue, but do you have any thoughts on whether a `glance()` method ought to report on the 

```
Pseudo likelihood ratio test = 0.07  on 2 df
```

note and, if so, what those entries should be called? Currently acceptable column names for glance outputs can be found [here](https://github.com/alexpghayes/modeltests/blob/master/data-raw/columns_glance.yaml), though we can extend that glossary if need be.

vincentarelbundock:
Unless we get expert advice to the contrary, it feels like the best way to go might be to store that test under the beautifully vague:

```r
statistic: Test statistic.
```

I force pushed that change so you can look at the whole picture.

simonpcouch:
Alright, I'm with it. Will fix merge conflicts in NEWS and then merge! :-)

PR #972, opened 2020-12-22
Fix bug in the output of the xyz list tidier

jiho:
```
A <- list(x = 1:5, y = 1:3, z = matrix(runif(5 * 3), nrow = 5))
df <- tidy(A)
glimpse(df)
# df$y is character instead of numeric
```

The fix uses a simpler code that avoids gather() and the change in the type of the y column that it brought.

It also adds tests to make sure the nature of the columns stays numeric in the future.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -72,6 +72,8 @@ TODO: sort out what happens to `glance.aov()`
 
 ## New tidiers, features and bugfixes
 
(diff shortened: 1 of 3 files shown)


jiho:
Closing this, I was based on an incorrect state of the master branch. Will reopen a new one if necessary.

PR #973, opened 2020-12-22
Fix bug in the output of the xyz list tidier

jiho:
```
A <- list(x = 1:5, y = 1:3, z = matrix(runif(5 * 3), nrow = 5))
df <- tidy(A)
glimpse(df)
# df$y is character instead of numeric
```

The fix uses a simpler code that avoids `pivot_longer()` and the change in the type of the y column that it brought.

It also adds tests to make sure the nature of the columns stays numeric in the future.


simonpcouch:
Well done and thanks for your patience on this! Just made some small edits, will merge after checks run. :-)

jiho:
Great, thanks!

Issue #645, opened 2019-03-09
Add augment.gam() method!


alexpghayes:

IndrajeetPatil:
Huh, `augment` currently seems to pick up `glm` method to do this-

``` r
library(mgcv)
#> Loading required package: nlme
#> This is mgcv 1.8-28. For overview type 'help("mgcv-package")'.

dat <- gamSim(1, n = 400, dist = "poisson", scale = .25)
#> Gu & Wahba 4 term additive model

b4 <- gam(y ~ s(x0) + s(x1) + s(x2) + s(x3),
  family = poisson,
  data = dat, method = "GACV.Cp", scale = -1
)

class(b4)
#> [1] "gam" "glm" "lm"

broom::augment(b4)
#> Warning in predict.gam(x, newdata, type = type.predict): not all required variables have been supplied in  newdata!
#> Error in eval(predvars, data, env): object 'x0' not found
```

<sup>Created on 2019-08-15 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

<details>

<summary>Session info</summary>

``` r
devtools::session_info()
#> - Session info ----------------------------------------------------------
#>  setting  value                       
#>  version  R version 3.6.1 (2019-07-05)
#>  os       Windows 10 x64              
#>  system   x86_64, mingw32             
#>  ui       RTerm                       
#>  language (EN)                        
#>  collate  English_United States.1252  
#>  ctype    English_United States.1252  
#>  tz       America/New_York            
#>  date     2019-08-15                  
#> 
#> - Packages --------------------------------------------------------------
#>  package     * version     date       lib
#>  assertthat    0.2.1       2019-03-21 [1]
#>  backports     1.1.4       2019-04-10 [1]
#>  broom         0.5.2.9002  2019-08-15 [1]
#>  callr         3.3.1       2019-07-18 [1]
#>  cli           1.1.0       2019-03-19 [1]
#>  crayon        1.3.4       2017-09-16 [1]
#>  desc          1.2.0       2019-04-03 [1]
#>  devtools      2.1.0       2019-07-06 [1]
#>  digest        0.6.20      2019-07-04 [1]
#>  dplyr         0.8.3       2019-07-04 [1]
#>  evaluate      0.14        2019-05-28 [1]
#>  fs            1.3.1       2019-05-06 [1]
#>  generics      0.0.2       2019-03-05 [1]
#>  glue          1.3.1       2019-03-12 [1]
[...]

PR #975, opened 2020-12-30
augment.gam

vincentarelbundock:
This PR addresses Issue https://github.com/tidymodels/broom/issues/645

Note that the the `.sigma` column is `NA`. This idiosyncratic behavior is already noted in the docs:

https://github.com/tidymodels/broom/blob/master/R/stats-lm-tidiers.R#L129


simonpcouch:
Hahaha I suppose I've been unforthcoming in how it is I'm formatting the attributions in NEWS--only citing contributors who aren't `"aut"` and I consider both issue filer and PR filer to be contributors. :-)

Thanks again for the high quality work! This looks great, only making a few small changes. Glad to see the helpers could do a lot of heavy lifting here.

vincentarelbundock:
ha! makes sense.

It's very motivating when maintainers are so responsive. But obviously, I understand that when I file a PR it entails work for you, so don't hesitate to let them sit for a while until it is convenient for you.

Cheers!

Issue #666, opened 2019-03-20
tidy.multinom() fails with error when the response is a matrix?


atyre2:
Fitting a multinomial model with no coefficients and 4 categories, I expect tidy() to produce a data.frame with
3 rows, one for each coefficient in the summary. Instead it fails with an error about nrow value < 0.

``` r
library(nnet)
library(broom)
response <- t(rmultinom(100, 1, c(0.1, 0.2, 0.3, 0.4)))
mmodel <- multinom(response~1)
#> # weights:  8 (3 variable)
#> initial  value 138.629436 
#> final  value 125.536505 
#> converged
summary(mmodel) # 3 coefficients
#> Call:
#> multinom(formula = response ~ 1)
#> 
#> Coefficients:
#>   (Intercept)
#> 2   0.6418557
#> 3   0.9555079
#> 4   1.5040790
#> 
#> Std. Errors:
#>   (Intercept)
#> 2   0.3906808
#> 3   0.3721044
#> 4   0.3496029
#> 
#> Residual Deviance: 251.073 
#> AIC: 257.073
tidy(mmodel) # breaks
#> Error in matrix(coef(s), byrow = FALSE, nrow = length(x$lev) - 1, dimnames = list(x$lev[-1], : invalid 'nrow' value (< 0)
```

Created on 2019-03-20 by the [reprex
package](http://reprex.tidyverse.org) (v0.2.0).

atyre2:
Update - maybe it has to do with the response variable being a matrix rather than a factor; I just made an example with a covariate and that also generates the same error.

```
X <- cbind(1, rnorm(100))
b0 <- c(0, .1, .2, .4)
b1 <- c(-1, 0, 1, 0)
p <- cbind(exp(X %*% c(b0[1], b1[1])),
      exp(X %*% c(b0[2], b1[2])),
      exp(X %*% c(b0[3], b1[3])),
      exp(X %*% c(b0[4], b1[4])))
psum <- rowSums(p)
p <- sweep(p, 1, psum, FUN = "/")
response <- apply(p, 1, function(x)rmultinom(1, 1, x))
mmodel <- multinom(t(response)~X[,2])
summary(mmodel)
```

brunaw:
Adressed in #618

mikemc:
@brunaw Can you clarify the status of the fix for this bug? The example posed by @atyre2 in https://github.com/tidymodels/broom/issues/666#issuecomment-475060146 still produces an error in the current development version.

[...]

PR #977, opened 2020-12-31
tidy.multinom accepts matrix response

vincentarelbundock:
This PR fixes Issue #666 

The problem was that the `multinom` object does not store a `x$lev` when the response is a matrix.


simonpcouch:
Bravo.🙂

hughjonesd:
Incidentally, did this come up on a reverse dependency check? My package got archived from CRAN because it failed a test using this....

Issue #161, opened 2016-09-15
broom for vars package on Vector Autoregressive Methods


Diego-MX:
Thank you for such an awesome package.  I've been using it for the last month and really want to exploit all of its potential.  
In my last project I found an interesting example that I can't get my head around.  So I want to put it out there, and see if I can get help from the community.  

The package `vars` gives models for timed vectors.  Here are links for the [docs](https://cran.r-project.org/web/packages/vars/vars.pdf) and [vignette](https://cran.r-project.org/web/packages/vars/vignettes/vars.pdf).  

The function `VAR` estimates models for the time-vector $y_t$ as follows:
$$ y_t = A_1 y_{t-1} + ... + A_p y_{t-p} $$    no LaTeX, =( 

But since $y_t$ contains several variables, its structure gets complicated. 

```
> library(vars)
> data("Canada")
> p1ct <- VAR(Canada, p=1, type="both")
> class(p1ct)
# "varest"
> names(p1ct)
# [1] "varresult"    "datamat"      "y"            "type"        
  [5] "p"            "K"            "obs"          "totobs"      
  [9] "restrictions" "call"  
```

The models are actually stored in `varresult` with the names of the variables. 

```
> class( p1ct$varresult$U )
# lm
> lm_U <- p1ct$varresult$U
```

For which I extracted the corresponding models in order to tidy them... and glance and augment them.  
There are two issues here: 
1. The main issue is that I couldn't get the fitted values back from the augmentation. 

```
compare_fitted <- cbind( lm_U$fitted.values, augment(lm_U)$.fitted )
compare_fitted
```
1. The second issue is minor.  When looking up about `augment`, I found that you can append the original data giving it as an argument.  In these models, the original data lost two rows because of the lagged observations.  So I got an error when I tried to recover it.  I was still able to get around it with `left_join` and `right_join`.  

In any case, my question is.  
a) How do the .fitted values get computed?  Would you know why I'm getting different ones?  
b) If I were to request methods for such `varest` class, or even better wanted to contribute in its development, how can I go about and do that?  I have been using R for a while, and am looking forward to taking the next step in contributing to a package.  

Cheers.

PR #979, opened 2021-01-01
new tidiers: vars package / varest objects

vincentarelbundock:
The goal of this PR is to address issue #161. It is marked as WIP because I have a couple questions a few notes.

Questions:

* The `vars` package does not define a `confint` method and does not report confidence intervals. Should the `tidy` method (a) include and document a `conf.int` argument and print an informative error if users set it to `TRUE`, or (b) exclude this argument from the function definition?
* One important information to report in `glance` is the lag order. Right now, I included a column called "order". Is that a good name? Should I make a PR to `modeltests`?

Note:

* `varest` objects include results for several levels. These are indicated in a `group` column produced by `tidy`.
* It is not clear to me what should be included in the `augment` output from a VAR model. For now, I do not define this function, but that might be something to consider in the future if an expert can help.
*  Right now, the `glance` test is commented out until `modeltests` includes and documents the column name.


simonpcouch:
As for your first bullet point, it's fine to exclude the `conf.*` level arguments and document the behavior in `@details`. :-)

Generally, when it comes to modeltests, feel free to file a PR there with the entries you need and link here. I'll merge there and broom can temporarily depend on the `Remotes:` version of modeltests until we push a release out to CRAN!

It's fine to not define an `augment` method, feel free to uncomment the `glance` method once the modeltests PR is merged.

vincentarelbundock:
Thanks for the info. 

I think there's an argument to be made for *including* `conf.*` and exiting with an error instead of sinking it in `...`. This would be more explicit, which I prefer, but I'm not dogmatic enough to argue if you prefer I remove the argument and mention this in docs.

simonpcouch:
I'm on board for a warning in that case!

vincentarelbundock:
Removing the [WIP] tag because I think this looks pretty good now. The remaining error should be fixed when `modeltests` is updated by this PR: https://github.com/alexpghayes/modeltests/pull/35

```r
== Failed tests ================================================================
-- Failure (test-vars.R:29:3): glance.vars -------------------------------------
length(unacceptable) == 0 is not TRUE

`actual`:   FALSE
`expected`: TRUE 
Output column names not in the column glossary: lag.order

[ FAIL 1 | WARN 25 | SKIP 3 | PASS 3886 ]
Error: Error: Test failures
Execution halted
```

simonpcouch:
Much appreciated! :-)

vincentarelbundock:
Ooof, this one wasn’t so clean. Thanks for the review!

Issue #780, opened 2019-11-12
Clarify studentized vs standardized residuals in augment.lm()


aloy:
For `lm` objects, could you add a `type.residuals` argument to allow for studentized residuals to be requested?

alexpghayes:
At the moment, if you (1) do not pass anything to `newdata`, the `.std.resid` column gives studentized residuals, rather than standardized residuals. We should probably document this better, and possibly allow users to get standardized residuals as well.

aloy:
Thanks! I agree that making this explicit in the documentation will fix this confusion point.

PR #980, opened 2021-01-01
[WIP] augment.lm includes studentized residuals

vincentarelbundock:
Marking as WIP because this probably needs discussion.

In [this comment](https://github.com/tidymodels/broom/issues/780#issuecomment-565620174), @alexpghayes writes that `augment` currently returns *studentized* residuals. I think that's a typo, because `augment_columns` actually adds *standardized* residuals produced by the `stats::rstandard` function: 

https://github.com/tidymodels/broom/blob/master/R/utilities.R#L206

In issue #780 @aloy recommends adding the `type.residuals` argument to allow `augment` to return *studentized* residuals. I don't think this is a good idea, because `stats::residuals.lm` accepts an argument of the same name with different inputs: "working", "response", "deviance", "pearson", "partial".

My proposal is trivial to implement (as demonstrated by this WIP PR): Add a new column called `.stu.resid` with the results produced by the `stats::rstudent` function.

I tried some simple benchmarks, and `rstudent` is usually faster to execute than the `rstandard` that `augment` already uses.

What do you all think?

### DIFF
--- a/R/utilities.R
+++ b/R/utilities.R
@@ -204,6 +204,7 @@ augment_columns <- function(x, data, newdata = NULL, type, type.predict = type,
   influence0 <- purrr::possibly(stats::influence, NULL)
   cooks.distance0 <- purrr::possibly(stats::cooks.distance, NULL)
   rstandard0 <- purrr::possibly(stats::rstandard, NULL)
+  rstudent0 <- purrr::possibly(stats::rstudent, NULL)
   predict0 <- purrr::possibly(stats::predict, NULL)
 
   # call predict with arguments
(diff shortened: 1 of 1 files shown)


alexpghayes:
I think adding new columns for each type of residual is confusing. In fact, I don't like the fact that we report both residuals and standardized residuals at the same time right now. I would prefer a `type.residual` argument that specifies that what goes into `.resid` and then no `.std.resid` column. However this is probably a breaking change. Other proposals on how to handle things without a breaking change or column proliferation appreciated.

Issue #992, opened 2021-02-12
tidy.multinom not working in broom 0.7.4


hughjonesd:
## The problem

In broom 0.7.2, the following worked fine. In broom 0.7.4, `tidy()` returns a 0-row data frame and `tidy(..., conf.int = TRUE)` throws an error.

## Reproducible example

```r
library(nnet)
set.seed(27101975)
  dfr <- data.frame(a = rnorm(100))
  dfr$y <- dfr$a + rnorm(100)

  mn <- nnet::multinom(I(y > 0) ~ a, dfr, trace = FALSE)
mn # coefficients "(Intercept)" and "a"
summary(mn) # fine
broom::tidy(mn)
broom::tidy(mn, conf.int = TRUE)
```

Useful info:

```
> packageVersion("nnet")
7.3.14
> R.version
                        
platform       x86_64-apple-darwin17.0     
arch           x86_64                      
os             darwin17.0                  
system         x86_64, darwin17.0          
status                                     
major          4                           
minor          0.3                         
year           2020                        
month          10                          
day            10                          
svn rev        79318                       
language       R                           
version.string R version 4.0.3 (2020-10-10)
nickname       Bunny-Wunnies Freak Out    
```

hughjonesd:
The problem seems to be these lines in `tidy.multinom`:
   
 ```
    co <- coef(s)
    coef <- matrix(co, byrow = FALSE, nrow = n_lev - 1, dimnames = list(row.names(co), 
        col_names))
```

The first just calls the generic since s is a summary.multinom object with no coef method. It returns the vector of statistics, with no row.names. Then later, `lapply(rownames(co), ...)` gets called and returns a 0-length list.

simonpcouch:
Thanks for the issue and diagnostic! Related to #977, I believe—will work on a fix over the weekend.

hughjonesd:
Incidentally, did this come up on a reverse dependency check? My package got archived from CRAN because it failed a test using this....

vincentarelbundock:
I apologize for this bug. `broom` did not have a test with a two-level dependent variable, and I did not anticipate that `nnet::coef.multinom` could return an object of different dimensions depending on the number of levels of the dependent variable.

I submitted a PR with one possible fix and a test using the reproducible example given above.

simonpcouch:
[...]

PR #993, opened 2021-02-13
nnet::multinom bugfix for 2-level y

vincentarelbundock:
Fix and test for bug reported in this issue: https://github.com/tidymodels/broom/issues/992

The problem was that when `y` has only two levels, `coef(fit)` returns a vector instead of a matrix, so we can't retrieve the row.names, which we use to name fill-in the `y.level` column in `tidy` output.

I decided to still include the `y.level` column here for consistency, and to assign it to "1".

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,6 +2,9 @@
 
 To be released as broom 0.7.5.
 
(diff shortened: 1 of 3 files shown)


simonpcouch:
Bravo! Looks good to me. Much appreciated.

PR #1002, opened 2021-03-10
Remove special handling of fixest `summary` objects.

### LINKED ISSUE #1001: fixest tidier ignores `se=` argument
## The problem

I noticed that the fixest tidier is ignoring the `se=` argument. This is happening because the hacky test I wrote in `is_fixest_summary` returns TRUE for some non-summary cases (as of fixest 0.8.3 and broom 0.7.5). Because this test thinks the object is a summary (possibly with the user's preferred standard error or dof settings), broom doesn't pass along the `se` argument or any others except `conf.int` and `conf.level`.

I'll try to submit a PR to better distinguish summary fixest objects from non-summarized ones.

https://github.com/tidymodels/broom/blob/1d6897515ecac6f072afe9c67138477e57d8eab9/R/fixest-tidiers.R#L74-L81

## Reproducible example

``` r
library(broom)

res = fixest::feols(Sepal.Length ~ Sepal.Width + Petal.Length , cluster=~Species, data=iris)
# Wrong result here:
broom:::is_fixest_summary(res)
#> [1] TRUE
broom:::is_fixest_summary(summary(res))
#> [1] TRUE

identical(tidy(res), tidy(res, se="standard"))
#> [1] TRUE

res2 = fixest::feols(Sepal.Length ~ Sepal.Width + Petal.Length, data=iris)
# Right result here:
broom:::is_fixest_summary(res2)
#> [1] FALSE
```

In this example, the `tidy_summary_fixest` code should have raised an error about being passed extra arguments. I'm not sure why it didn't.

https://github.com/tidymodels/broom/blob/1d6897515ecac6f072afe9c67138477e57d8eab9/R/fixest-tidiers.R#L84-L92

karldw:
fixest, since v0.8.1, remembers the settings passed to `summary`, so `tidy.fixest` can call `summary` again without losing any settings that were passed in the first `summary` call. That means we don't have to handle pre-summarized objects differently.

Fixes #1001 (see reprex in that issue).

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -552,7 +552,7 @@ Suggests:
     emmeans,
     epiR,
     ergm (>= 3.10.4),
-    fixest (>= 0.5.0),
+    fixest (>= 0.8.1),
     gam (>= 1.15),
     gamlss,
     gamlss.data,
--- a/NEWS.md
+++ b/NEWS.md
@@ -15,7 +15,7 @@ variable has only two levels (`#993` by `@vincentarelbundock` and `@hughjonesd`)
 
 # broom 0.7.4
 
-broom 0.7.4 introduces tidier support for a number of new model objects and
+broom 0.7.4 introduces tidier support for a number of new model objects and 
 improves functionality of many existing tidiers!
 
 #### New Tidiers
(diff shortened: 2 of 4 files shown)


simonpcouch:
Bravo! This looks great.

Pushing a quick commit to clear up the diff in NEWS.md and then will merge. Thank you. :-)

karldw:
Thank you!

PR #1005, opened 2021-03-21
Revert to intentional typo in docs

LukasWallrich:
The param_unsed_dots template intends to warn the user what happens when they enter a typo. Commit 71dc8d9b5c4c69330bd3c0263334da6c8385ad50 fixed that typo, which turned this documentation entry into something that's very confusing / misleading. Here, I propose to re-introduce that typo.


simonpcouch:
Thanks so much for catching this! re`devtools::document()`ed all of the help files, will merge once checks pass. :-)

PR #1013, opened 2021-04-12
Adds the `exponentiate=` argument to `tidy.gam()`

ddsjoberg:
PR adds the `exponentiate=` argument to `tidy.gam()`. 

- The argument only applies when `parametric = TRUE` (similar to `conf.int=`). 
- If user requests exponentiated coefs when `parametric = FALSE`, they are notified the argument has been ignored (just like what happens with  `conf.int=`).


alexpghayes:
Thank you!

Issue #1023, opened 2021-05-20
tidy.crr()- CI exponentiated with argument set to FALSE


leejasme:
## The problem

The confidence intervals (CI) when using the tidy.crr() function for the cmprsk::crr() models shows exponentiated CI when exponentiate = F.

Pull request for this to come shortly. 

## Reproducible example
``` r
set.seed(10)
lrf_time <- rexp(100) #time to loco-regional failure (lrf)
lrf_event <- sample(0:2, 100, replace = TRUE) 
trt <- sample(0:1, 100, replace = TRUE)
strt <- sample(1:2, 100, replace = TRUE)
x <- cmprsk::crr(lrf_time, lrf_event, cbind(trt, strt))
# CIs from the cmprsk::summary.crr 
cmprsk::summary.crr(x, conf.int = 0.95)
#> Competing Risks Regression
#> 
#> Call:
#> cmprsk::crr(ftime = lrf_time, fstatus = lrf_event, cov1 = cbind(trt, 
#>     strt))
#> 
#>        coef exp(coef) se(coef)      z p-value
#> trt  -0.241     0.786    0.332 -0.726    0.47
#> strt -0.273     0.761    0.330 -0.829    0.41
#> 
#>      exp(coef) exp(-coef)  2.5% 97.5%
#> trt      0.786       1.27 0.410  1.51
#> strt     0.761       1.31 0.399  1.45
#> 
#> Num. cases = 100
#> Pseudo Log-likelihood = -135 
#> Pseudo likelihood ratio test = 1.06  on 2 df,

# CI from broom::tidy.crr
broom::tidy(x, exponentiate = F, conf.int = T)
#> # A tibble: 2 x 7
#>   term  estimate std.error statistic p.value conf.low conf.high
#>   <chr>    <dbl>     <dbl>     <dbl>   <dbl>    <dbl>     <dbl>
#> 1 trt     -0.241     0.332    -0.726    0.47    0.410      1.51
#> 2 strt    -0.273     0.330    -0.829    0.41    0.399      1.45
```

<sup>Created on 2021-05-20 by the [reprex package](https://reprex.tidyverse.org) (v2.0.0)</sup>

PR #1024, opened 2021-05-20
(Issue #1023) tidy.crr take log(CI) when exponentiate = F

leejasme:
PR to fix issue #1023 : the confidence intervals for tidy.crr() which displays exponentiated CI when `exponentiate = F`

- summary.crr provides exponentiated CI 
- changed the default to show the log(CI) from summary.crr instead to match the default of `exponentiate = F`


alexpghayes:
Given the estimate is named `estimate_exp()` it seems like we're suggesting via the column name that the estimate is always exponentiated. Is this reasonable for these models?

If we make exponentiation optional (see also: the exponentiation idiom), then we should change the column names of the output. 

In any case, the fact that the current code suggests that you can control exponentiation but you cannot is problematic.

leejasme:
I’ve only ever reported the exponentiated estimates and CI in the ~1 year since I’ve started using crr, not sure if others find the original estimates and CI useful. 
I think it’s good to have that option since  exponentiate = F seems to be default in most cases , I would expect that I need to pass in an exponentiate = T when I’m using tidy.crr

ddsjoberg:
Thank you @leejasme for submitting this fix, and @alexpghayes for reviewing! I wanted to add a few comments to this update.

The default summary function for a crr model returns both the regression coef and the exponentiated version. While it is indeed common to exponentiate the coef to obtain the sub-hazard ratio (just like in Cox regression to get the hazard ratio, or in logistic regression to get the odds ratio), there are many situations when you need the coef and not the hazard ratio, e.g. when there is an interaction and the HR is no longer interpretable. I _strongly_ suggest to keep the default `exponentiate = FALSE` as this is in alignment with the many other linear regression tidiers and likely what users would expect the default behavior to be. Moreover, it is possible to build competing risk regression models using the survival package, and the resulting  tibble from that tidier also defaults to `exponentiate = FALSE`.

Thank you again both!

leejasme:
Hi @alexpghayes , I updated the `new_names` in `ci` to reflect the actual values. The `estimate_exp` is now `estimate` and `estimate_neg_exp` is `estimate_neg`.  Will this work? 
Thanks @ddsjoberg  for helping me with this !

ddsjoberg:
Hey @leejasme! I think @alexpghayes retired from broom dev, and @simonpcouch is now the broom CEO?

alexpghayes:
Simon is now the maintainer :)

ddsjoberg:
Thank you for your service @alexpghayes ! I would think broom is one of the most difficult packages to maintain with the constant changes in the dependencies!

leejasme:
thanks @ddsjoberg , and @alexpghayes for all your work in maintaining this incredibly useful package!

simonpcouch:
Hey yall! Thanks for catching this and the clear explanation. I appreciate your patience on this—in the middle of a move at the moment.

I agree that this is something we should make sure we get right! The `estimate_exp` and `estimate_neg_exp` aren't user-facing, as they're selected off after the `as_tidy_tibble()` step, so I think it's okay to keep that `new_names` argument as is in the current source! I'm on board for the `log(CI)` change, though, so will commit that quickly and merge once checks look okay!

Thanks again, yall.

Issue #1031, opened 2021-06-21
Unintended labeling consequence in augment_newdata()?


rpruim:
This line in the unexported `augment_newdata()`

```r
    if (!is.null(resp) && is.numeric(resp)) {
        df$.resid <- (resp - df$.fitted) %>% unname()
    }
```
has the side effect that `.resid` inherits attributes from `resp`, the response variable.  This includes the "label" attribute, which is used by packages like {labelled} and {declared} to provide labels for variables.  But the label is typically not an accurate label for the residuals.

Here are some examples.

``` r
suppressPackageStartupMessages(library("ggformula"))
suppressPackageStartupMessages(library("labelled"))
suppressPackageStartupMessages(library("declared"))
suppressPackageStartupMessages(library("tibble"))

# we can add labels using either {labelled} or {declared}

some_data <- 
  tibble(x = 1:10, y1 = x + rnorm(10), y2 = declared(y1, label = "declared y")) %>%
  labelled::set_variable_labels(y1 = "labelled y")

lapply(some_data, attributes)
#> $x
#> NULL
#> 
#> $y1
#> $y1$label
#> [1] "labelled y"
#> 
#> 
#> $y2
#> $y2$class
#> [1] "declared" "numeric" 
#> 
#> $y2$label
#> [1] "declared y"

# the labels persist in .resid produced by broom::augment()

lm(y1 ~ x, data = some_data) %>%
  broom::augment() %>%
  gf_point(.resid ~ x)
```

![](https://i.imgur.com/eSdyjUD.png)

``` r
lm(y2 ~ x, data = some_data) %>%
  broom::augment() %>%
  gf_point(.resid ~ x)
```

![](https://i.imgur.com/5n4VNhJ.png)

``` r
# similar things happen with labels of other quantities as well:

lm(y1 ~ x, data = some_data) %>%
  broom::augment() %>%
  lapply(function(x) attr(x, 'label'))
#> $y1
#> [1] "labelled y"
#> 
#> $x
#> NULL
#> 
#> $.fitted
#> NULL
#> 
#> $.resid
#> [1] "labelled y"
#> 
#> $.std.resid
#> [1] "labelled y"
#> 
#> $.hat
#> NULL
#> 
#> $.sigma
#> NULL
#> 
#> $.cooksd
#> [1] "labelled y"

lm(y2 ~ x, data = some_data) %>%
  broom::augment() %>%
  lapply(function(x) attr(x, 'label'))
#> $y2
[...]

PR #1032, opened 2021-06-21
set `resid` `"label"` attr to NULL

simonpcouch:
Related to #1031. Will check reverse dependencies before merging.

### DIFF
--- a/R/utilities.R
+++ b/R/utilities.R
@@ -437,6 +437,7 @@ augment_newdata <- function(x, data, newdata, .se_fit, interval = NULL, ...) {
 
   if (!is.null(resp) && is.numeric(resp)) {
     df$.resid <- (resp - df$.fitted) %>% unname() 
+    attr(df$.resid, "label") <- NULL
   }
 
   df


alexpghayes:
Instead of setting a single attribute to `NULL` it might be nice to strip all attributes here instead.

rpruim:
If you are onboard with this, we should fix `.std.resid`, `.cooksd` (and perhaps other things for different types of models).

I didn't want to propose a particular fix without having a better idea whether this would have negative consequences elsewhere, and I don't know enough to weigh in on removing only the labels or all of the attributes.

alexpghayes:
Basically `broom` assumes that columns constructed in `augment()` are vanilla vectors. When residuals have classes or attributes beyond this it will cause unexpected behaviors. To be as transparent as possible I think the approach to take is to attempt to coerce everything to vector, stripping all attributes, and then if this fails, error informatively.

But essentially this is an upstream issue in that there is a class/type/whatever that allows methods (in this case subtraction) that it shouldn't. We don't want to special case to handle this sort of poor etiquette because otherwise the maintenance burden explodes.

simonpcouch:
Yeah, that's a fair evaluation Alex--thanks for the input. Since the `"label"` attribute is introduced from `labelled`, I think we'll expect dependent packages to handle those outputs.

larmarange:
> Basically `broom` assumes that columns constructed in `augment()` are vanilla vectors. When residuals have classes or attributes beyond this it will cause unexpected behaviors. To be as transparent as possible I think the approach to take is to attempt to coerce everything to vector, stripping all attributes, and then if this fails, error informatively.
> 
> But essentially this is an upstream issue in that there is a class/type/whatever that allows methods (in this case subtraction) that it shouldn't. We don't want to special case to handle this sort of poor etiquette because otherwise the maintenance burden explodes.

A quick reply. It is important to distinct variable labels and value labels introduced by `haven` (and also implemented in `labelled`).

When value labels are defined, it will convert vectors to a dedicated class (`haven_labelled`). For a specific class, it is possible to declare a relevant method.

However, variable labels are just a simple attribute attached to any vanilla vector, and it doesn't change the nature or the class of the vector, but simple add a `"label"` attribute to the vector using `base::attr()`. **In this discussion, the issue was about a variable label.** This label act like any other attributes, like names of vector elements.

I'm not sure about R rules regarding attributes and arithmetic.

According to the examples below (look both at names and at the label attributes), it seems that attributes are always kept. If you combine two vectors with different values of the same attribute, only the first value is kept.

I do not see any mechanism to state when and when not attributes should be kept in arithmetic operations on vectors.

``` r
x <- c(a = 1, b = 2, c = 3, d = 4)
attr(x, "label") <- "variable X"

y <- 4:1
attr(y, "label") <- "variable Y"

x
#> a b c d 
#> 1 2 3 4 
#> attr(,"label")
#> [1] "variable X"
attributes(x)
#> $names
#> [1] "a" "b" "c" "d"
#> 
#> $label
#> [1] "variable X"

x - 2
#>  a  b  c  d 
#> -1  0  1  2 
#> attr(,"label")
#> [1] "variable X"

x + 1:4
#> a b c d 
#> 2 4 6 8 
#> attr(,"label")
#> [1] "variable X"

x + y
#> a b c d 
#> 5 5 5 5 
#> attr(,"label")
#> [1] "variable X"

1:4 + x
#> a b c d 
#> 2 4 6 8 
#> attr(,"label")
#> [1] "variable X"

y + x
#> a b c d 
#> 5 5 5 5 
#> attr(,"label")
#> [1] "variable Y"
```

<sup>Created on 2021-06-22 by the [reprex package](https://reprex.tidyverse.org) (v2.0.0)</sup>

larmarange:
Additional quick comment: in set resid, you already have a call to `unname()`:  ` df$.resid <- (resp - df$.fitted) %>% unname() `

You may consider a more generic:

```r
df$.resid <- (resp - df$.fitted)
attributes(df$.resid) <- NULL
```

forcing a vanilla result

simonpcouch:
For posterity, I agree that an approach a la https://github.com/tidymodels/broom/pull/1032#issuecomment-866249604 is likely the way to go. However, this would likely be a breaking change for several dependencies and we don't have the maintenance capacity to handle that gracefully for now.

As for your first comment, @larmarange, this sounds like a context where the implementing package ought to append a class to its output so it can define the behavior of its operators when needed, as you discuss.

larmarange:
A final thought.

Variable labels are just a piece of information (a long variable name) attached to any vector type (numeric, factor, Date, Posixtc, character, hms, other classes...) that does not change the nature of the vector. In that sense, there is no reason to change the class (the nature) of the vector. This is exactly the role of attributes: attaching a piece of information to an object.

There is probably a discussion around how attributes should be kept or not in arithmetic operations, but this is a general discussion to have with the core R team.

> However, this would likely be a breaking change for several dependencies 

Just by curiosity, in that very specific case, where there is a `is.numeric()` condition just before, do you see any case where it will be problematic to no propagate attributes for this specific computation?

simonpcouch:
> in that very specific situation, where there is a is.numeric() condition just before, do you see any case where it will be problematic to not propagate attributes for this specific computation?

Yes, as demonstrated in #1031.

rpruim:
I think I'd also be happy with https://github.com/tidymodels/broom/pull/1032#issuecomment-866249604.  Are there any potential downsides?  That is, is there any situation where we want the residuals, Cook's distance, etc. to inherit attributes from the response variable?

This solution also seems to fit with

> Basically broom assumes that columns constructed in augment() are vanilla vectors. 

But instead of assuming, it is enforcing.  This could be documented, so it is clear that the results will be vanilla.

alexpghayes:
Oh I like this!

> ```r
> df$.resid <- (resp - df$.fitted)
> attributes(df$.resid) <- NULL
> ```

Breakages are probably worth handling, I can try to pitch in if there are a bunch if someone makes a PR with `revdepcheck` results.

rpruim:
We should handle the other new columns at the same time (Cook's distance, etc.). I'd have to look at the code to see if we can get them all at once by using 

```r
attributes(resp) <- NULL
```
or weather some of these don't get computed directly from `resp` like `.resid` does.

Issue #1058, opened 2021-10-13
warn_on_glm2 breaks with method function


gravesti:
## The problem
The `warn_on_glm2` function causes an error for a `stats::glm` object where the method argument has been given a function. From the glm docs, 
for the method argument:
>the method to be used in fitting the model. The default method "glm.fit" uses iteratively reweighted least squares (IWLS): the alternative "model.frame" returns the model frame and does no fitting.
>
>User-supplied fitting functions can be supplied either as a function or a character string naming a function, with a function which takes the same arguments as glm.fit. If specified as a character string it is looked up from within the stats namespace.

and the method vaue:

> the name of the fitter function used (when provided as a character string to glm()) or the fitter function (when provided as that).

The trouble seems to be here, where we test equality between a function and a character.
https://github.com/tidymodels/broom/blob/0f4c1caa8d7c3f7162849c3d892e522a913ed966/R/stats-glm-tidiers.R#L166

## Reproducible example

```r
## Based on the example from the glm documentation
## Dobson (1990) Page 93: Randomized Controlled Trial :
counts <- c(18,17,15,20,10,20,25,13,12)
outcome <- gl(3,1,9)
treatment <- gl(3,3)
glm.D93 <- glm(counts ~ outcome + treatment, family = poisson(),
               method = glm.fit)

broom::tidy(glm.D93)
#> Error in x$method == "glm.fit2": comparison (1) is possible only for atomic and list types
```

PR #1059, opened 2021-10-15
warn_on_glm2 check for character

gravesti:
Check x$method is a character before testing if it equals "glm.fit2" otherwise functions will cause errors

A possible fix for #1058


simonpcouch:
Thanks again. :-)

PR #1060, opened 2021-10-24
Added Wald confidence intervals for glm - resolves #989

### LINKED ISSUE #989: Specify the confint method in tidy.glm()
Adding the possibility to specify the desired `confint` method via an argument to `tidy.glm()` would be terrific. 
My user case is to calculate Wald confidence intervals (`confint.default()`) instead of the default profile likelihood CIs for `glm` objects (`confint.glm()`).

See also the last 3 messages in #703.

lcgodoy:
For this pull request, the feature requested in issue #989 was included in the package.
The `conf.method` parameter was included in the `tidy.glm` method. Its possible values are, the default, "plik" for computing the profile likelihood CI or "wald" for the Wald CI. In order the make it work consistently, the function `broom_confint_terms` had to be modified. These changes have not impacted other methods that depend on `broom_confint_terms`.


simonpcouch:
Hey there, thanks for the PR!

Apologies on the change of mind here, but I no longer feel that moving forward with this feature is the right move. Given where this package is at in its lifecycle and my own (in)ability to maintain this functionality (and the addition in scope it otherwise welcomes) well, I don't feel that I'm comfortable with adding features to the bread-and-butter `stats` tidiers.

alexpghayes:
@simonpcouch I'd be willing to take on maintenance duties for this if that changes your opinion here. This change really should only be local to the `tidy.glm()` and we could implement it without touching `broom_confint_terms` at all, which should keep the maintenance burden low and avoid spooky behavior.

simonpcouch:
@alexpghayes I'm for it! Leaving `broom_confint_terms` alone sounds like the right move to me as well.

alexpghayes:
@lcgodoy I'll add a quick code review momentarily.

R/stats-glm-tidiers.R
```diff
@@ -5,13 +5,20 @@
 #' @template param_confint
 #' @template param_exponentiate
 #' @template param_unused_dots
+#' @param conf.method Controls whether stats::confint.default (Wald method)
```
alexpghayes, line 8: Here I would emphasis the choice between estimators/methods rather than the code underlying those methods. i.e. Wald method is more important to know about than `stats::confint.default()`. Additionally, it would be nice to mention what `summary.glm()` does/what you get if you call `summary()` on a `glm` object with default arguments, since that's unclear from the method names themselves.

R/stats-glm-tidiers.R
```diff
@@ -5,13 +5,20 @@
...
+#'   or stats::confint (Profile likelihood method) will be used to compute
+#'   the coefficients confidence intervals. The former is typically faster.
+#'   The allowed types are `"plik"` for profile likelihood or `"wald"`
```
alexpghayes, line 11: Let's change `plik` to `profile`. Whole words are the preferred style in `broom`.

R/utilities.R
```diff
@@ -465,11 +465,15 @@ broom_confint <- function(x, ...) {
 }
 
 # this version adds a terms column
-broom_confint_terms <- function(x, ...) {
+broom_confint_terms <- function(x, conf.method = "plik", ...) {
```
alexpghayes, line 468: Instead of making this change in `broom_confint_terms`, let's relocate this code into the `tidy.glm()` method.

alexpghayes (changes requested):
Some comments about naming and code organization. If possible, it would also be nice to remove the `.Rd` changes from the pull request for a cleaner git history.

DanChaltiel:
Hi everyone,

I just added the issue https://github.com/tidymodels/broom/issues/1074 which is highly related to this PR.

There, I suggest that when this feature is accepted, it should become the default behaviour of `tidy.glm()`.

This might lead to minor breaking changes but it would correct the mismatch between p-values and confidence intervals, which I find important.

R/utilities.R
alexpghayes: Revisiting and I agree with my prior point here -- this change should be local to `glm` and `lm` methods, rather than occurring in a general helper function.

alexpghayes (changes requested):

snhansen:
Hi,
I'm also very interested in the ability of specifying confidence interval method. What's the status of this request? Can we see it implemented soon?

simonpcouch:
As we've been unable to bring together the development + maintenance energy for this changes in the last two years, I'm going to go ahead and close. Thanks all for the efforts.🌸

PR #1073, opened 2022-01-11
Update title_desc_tidy_list.R

PursuitOfDataScience:
Slightly modified the documentation so that it is more readable. Thanks.


simonpcouch:
Great, thanks! Just pushed some small edits and ran `devtools::document()`--will merge on passing checks.

Issue #1018, opened 2021-05-01
`glance()` error when using a `fixest` object with only FEs


arcruz0:
Thanks for the great package. I'm having trouble with `glance()` when using a `fixest` model with only FEs (no regular covariates). Please see the reproducible example below.

```r
# R 4.0.5 / Windows 10 Pro (64-bit)
library(broom)   # tested with 0.7.6 (CRAN) and 0.7.6.9001 (GH)
library(fixest)  # tested with 0.8.4 (CRAN)

ex_fe <- feols(Sepal.Length ~ 1 | Species, data = iris)
```

`summary()` from `fixest` works as expected:
```r
summary(ex_fe)
# OLS estimation, Dep. Var.: Sepal.Length
# Observations: 150 
# Fixed-effects: Species: 3
# RMSE: 0.509616   Adj. R2: 0.613518
```

However, `broom::glance()` returns an error. I think this also causes `modelsummary::modelsummary()` to fail (which is how I found the error):

```r
broom::glance(ex_fe)
# Error in eval_tidy(xs[[j]], mask) : object 'sigma2' not found

modelsummary::modelsummary(ex_fe)
# Error: $ operator is invalid for atomic vectors
```

I believe the expected behavior would be to return a df with missing values when needed, such as what happens with `lm` objects:

```r
ex_lm <- lm(Sepal.Length ~ 1, data = iris)

broom::glance(ex_lm)
# A tibble: 1 x 12
# r.squared adj.r.squared sigma statistic p.value    df
# <dbl>         <dbl> <dbl>     <dbl>   <dbl> <dbl>
#   1         0             0 0.828        NA      NA    NA
# # ... with 6 more variables: logLik <dbl>, AIC <dbl>,
# #   BIC <dbl>, deviance <dbl>, df.residual <int>,
# #   nobs <int>
```

vincentarelbundock:
I submitted a Pull Request to fix this. https://github.com/tidymodels/broom/pull/1088

PR #1088, opened 2022-03-19
Bug #1018: glance.fixest catches error

vincentarelbundock:
https://github.com/tidymodels/broom/issues/1018

When a `fixest` model includes only fixed effects and no regressors `sigma2` does not exist and `glance.fixest()` errors. I use `tryCatch` to return `NA_real_` instead. I also added a test.


simonpcouch:
Thanks for all of your work in these recent PRs! I'll be getting back up to speed on tidymodels Issues/PRs in early April and will return to these PRs then. :)

vincentarelbundock:
No rush at all. Boundaries are important ;)

PR #1089, opened 2022-03-24
Switch to tibble()

grantmcdermott:
I know these data.frame tidiers are going to be deprecated soon. But currently they trigger an unnecessary double warning because of the old `data_frame()` calls. This is just a quick patch to avoid those messages in the meantime.


alexpghayes:
Thanks!

PR #1095, opened 2022-04-05
fix builds + pare down Actions

simonpcouch:
Starting off with a super simplified version of the current Action and seeing if we can trim off some of the excess re: package management. Guessing some amount of this will need to be re-introduced to get `sf` and its dependencies situated.


simonpcouch:
Green checks! Huzzah!

Issue #1078, opened 2022-02-06
Methods for class "svyglm"


mlaviolet:
Why does the `svyglm` class have methods for `glance` and `tidy` but not `augment`? The warning gives the impression that no methods are available for this class but is only displayed with `augment`.

```r
library(broom)
library(survey)
#> Loading required package: grid
#> Loading required package: Matrix
#> Loading required package: survival
#> 
#> Attaching package: 'survey'
#> The following object is masked from 'package:graphics':
#> 
#>     dotchart

data(api)
dstrat <- svydesign(id = ~1, strata = ~stype, weights = ~pw, 
                    data = apistrat, fpc = ~fpc)
svymodel <- svyglm(api00 ~ ell + meals + mobility, design = dstrat)

glance(svymodel)
#> # A tibble: 1 x 6
#>   null.deviance df.null   AIC      BIC deviance df.residual
#>           <dbl>   <int> <dbl>    <dbl>    <dbl>       <dbl>
#> 1      3022927.     199  6.13 1029242. 1029221.         194
tidy(svymodel)
#> # A tibble: 4 x 5
#>   term        estimate std.error statistic   p.value
#>   <chr>          <dbl>     <dbl>     <dbl>     <dbl>
#> 1 (Intercept)  821.       10.1      81.5   5.59e-152
#> 2 ell           -0.481     0.392    -1.23  2.22e-  1
#> 3 meals         -3.14      0.284   -11.1   2.25e- 22
#> 4 mobility       0.226     0.393     0.574 5.67e-  1
augment(svymodel)
#> Warning: Tidiers for objects of class svyglm are not maintained by the broom
#> team, and are only supported through the glm tidier method. Please be cautious
#> in interpreting and reporting broom output.
#> # A tibble: 200 x 11
#>    api00   ell meals mobility `(weights)` .fitted   .resid .std.resid    .hat
#>    <int> <int> <int>    <int>       <dbl> <svystat>  <dbl>      <dbl>   <dbl>
#>  1   840    25    33       11        1.43 707.6848  158.       2.21   0.0138 
#>  2   516    77    98       26        1.43 481.8802   40.8      0.584  0.0581 
#>  3   531    23    64       17        1.43 612.6127  -97.5     -1.36   0.0120 
# ...
#> # ... with 190 more rows, and 2 more variables: .sigma <dbl>, .cooksd <dbl>

?broom::glance.svyglm
?broom::tidy.svyglm
?broom::augment.svyglm
[...]

PR #1098, opened 2022-04-11
clarify warning for un-maintained tidiers

simonpcouch:
re: #1078, workshop the warning for tidiers only supported via `lm`/`glm` to be clear that some tidiers for a class may be "officially" maintained while others of the same class only receive support via dispatch. What used to read:

```
Warning message:
Tidiers for objects of class svyglm are not maintained by the broom team, and are only 
supported through the glm tidier method. Please be cautious in interpreting and 
reporting broom output. 
```

now reads:

```
Warning message:
The `augment()` method for objects of class svyglm is not maintained by the broom 
team, and is only supported through the glm tidier method. Please be cautious in 
interpreting and reporting broom output. 
```


simonpcouch:
Closes #1078 :)

PR #1106, opened 2022-06-06
Support for car::lht and better term handling

### LINKED ISSUE #1090: Improve anova tidiers
The current tidiers for anova objects could do with some improving IMO. Consider the following example, which essentially replicates the example offered by the `?tidy.anova` help file:

``` r
library(broom)

a <- lm(mpg ~ wt + qsec + disp, mtcars)
b <- lm(mpg ~ wt + qsec, mtcars)
(ab <- anova(a, b))
#> Analysis of Variance Table
#> 
#> Model 1: mpg ~ wt + qsec + disp
#> Model 2: mpg ~ wt + qsec
#>   Res.Df    RSS Df  Sum of Sq     F Pr(>F)
#> 1     28 195.46                           
#> 2     29 195.46 -1 -0.0010239 1e-04 0.9904
tidy(ab)
#> # A tibble: 2 × 6
#>   res.df   rss    df    sumsq statistic p.value
#>    <dbl> <dbl> <dbl>    <dbl>     <dbl>   <dbl>
#> 1     28  195.    NA NA       NA         NA    
#> 2     29  195.    -1 -0.00102  0.000147   0.990
glance(ab)
#> Warning: Data frame tidiers are deprecated and will be removed in an upcoming
#> release of broom.
#> Warning: `data_frame()` was deprecated in tibble 1.1.0.
#> Please use `tibble()` instead.
#> This warning is displayed once every 8 hours.
#> Call `lifecycle::last_lifecycle_warnings()` to see where this warning was generated.
#> # A tibble: 1 × 4
#>    nrow  ncol complete.obs na.fraction
#>   <int> <int>        <int>       <dbl>
#> 1     2     6            1       0.333
```

<sup>Created on 2022-03-24 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

I'll take the `tidy` and `glance` methods one by one:

### 1. tidy.anova

The current return object makes it impossible to ascertain what the original ANOVA test was. We have no meaningful information about the differences between the comparison models. There are a couple ways to approach this, but I think the simplest is to have a dedicated "model" column that reflects the underlying terms in a character string. So, add something like this internally

``` r
(comp_mods = sub(".*: ", "", strsplit(attr(ab, "heading")[[2]], "\n")[[1]]))
#> [1] "mpg ~ wt + qsec + disp" "mpg ~ wt + qsec"

cbind(tidy(ab), model = comp_mods)
#>   res.df      rss df        sumsq    statistic   p.value                  model
#> 1     28 195.4626 NA           NA           NA        NA mpg ~ wt + qsec + disp
#> 2     29 195.4636 -1 -0.001023935 0.0001466785 0.9904229        mpg ~ wt + qsec
```

(Another option is just to use the generic "term" column, since then you don't need to add a new glossary term to pass the modeltests check.)

### 2. glance.anova

Here the problem arises due to the (incorrect) data.frame method passing. I know that this behaviour is going to be fully deprecated soon. But perhaps we should add a dedicated `anova.glance` method that returns an empty data.frame to head this off explicitly?

grantmcdermott:
(Mostly) fixes #1090.

I was very tempted to create a new, dedicated `glance.anova` method and pull some columns that are currently returned as part of the `tidy.anova` object. (Most obviously, the "df.residual" and "rss" columns.) But for now I'll stick to a more conservative fix that just supports `car::linearHypothesis` and returns more information about the model contrasts. OTOH, the fact that `glance.anova` currently still returns a not very sensible data frame is probably something that needs to be fixed at some point.

Some examples:

``` r
devtools::load_all("~/Documents/Projects/broom")
#> ℹ Loading broom

library(broom)

a <- lm(mpg ~ wt + qsec + disp, mtcars)
b <- lm(mpg ~ wt + qsec, mtcars)

ab <- anova(a, b)

# new change: term column shows comparison models
tidy(ab)
#> # A tibble: 2 × 7
#>   term                   df.residual   rss    df    sumsq statistic p.value
#>   <chr>                        <dbl> <dbl> <dbl>    <dbl>     <dbl>   <dbl>
#> 1 mpg ~ wt + qsec + disp          28  195.    NA NA       NA         NA    
#> 2 mpg ~ wt + qsec                 29  195.    -1 -0.00102  0.000147   0.990

library(car)
#> Loading required package: carData
#> 
#> Attaching package: 'car'
#> The following object is masked from 'package:broom':
#> 
#>     recode
tidy(lht(a, "wt = disp"))
#> # A tibble: 1 × 10
#>   term     null.value estimate std.error statistic p.value df.residual   rss    df
#>   <chr>         <dbl>    <dbl>     <dbl>     <dbl>   <dbl>       <dbl> <dbl> <dbl>
#> 1 wt -disp          0    -5.03      1.23      16.6 3.39e-4          28  195.     1
#> # … with 1 more variable: sumsq <dbl>
```

<sup>Created on 2022-06-06 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>


simonpcouch:
Feel free to ignore the pkgdown checks—that machinery needs updating.🌝🌚

grantmcdermott:
Thanks Si!

I can add the `glance.anova` method to this PR if you'd prefer. Here's what that would entail:

- Dropping any "df.residual" and "rss" columns from the returned `tidy.anova` object.
- Porting those columns to the `glance` method instead. (And maybe renaming "rss" to "deviance" to be consistent with other glance methods?)
- In some default cases&mdash;e.g. `anova(lm(mpg ~ wt, mtcars)`&mdash;the glance emthod would just be an empty data frame, since the return object doesn't produce appropriate glance-like columns.

Lmk your thoughts and I'll try to submit ASAP.

simonpcouch:
> Porting those columns to the glance method instead. (And maybe renaming "rss" to "deviance" to be consistent with other glance methods?)

On board!

> In some default cases—e.g. anova(lm(mpg ~ wt, mtcars))—the glance method would just be an empty data frame, since the return object doesn't produce appropriate glance-like columns.

On board!

> Dropping any "df.residual" and "rss" columns from the returned tidy.anova object.

I think I'd hold off on this. These columns have been around, at least by position, since some of the [first commits](https://github.com/tidymodels/broom/blob/f04a7fa8200af378d311765d827e910f2ff73603/R/tidy-stats.R) to broom. I see the argument for why they shouldn't be there, but I'd imagine this would affect a good few reverse dependencies.

grantmcdermott:
just to let you know i haven't forgotten this... need to get grading done first, though :'-|

simonpcouch:
@grantmcdermott No rush. :)

grantmcdermott:
Thanks for bearing with me @simonpcouch. I think these last few changes should do it.

I added the following note to the `glance.anova()` help documentation.

```r
#' Note that the output of `glance.anova()` will vary depending on the initializing 
#' anova call. In some cases, it will just return an empty data frame. In other 
#' cases, `glance.anova()` may return columns that are also common to
#' `tidy.anova()`. This is partly to preserve backwards compatibility with early
#' versions of `broom`, but also because the underlying anova model yields 
#' components that could reasonably be interpreted as goodness-of-fit summaries
#' too.
```

Example:

``` r
devtools::load_all("~/Documents/Projects/broom")
#> ℹ Loading broom

a <- lm(mpg ~ wt + qsec + disp, mtcars)
b <- lm(mpg ~ wt + qsec, mtcars)

ab <- anova(a, b)

glance(ab)
#> # A tibble: 1 × 2
#>   deviance df.residual
#>      <dbl>       <dbl>
#> 1     195.          29

## Example where glance returns an empty DF
glance(anova(a))
#> # A tibble: 0 × 0
```

<sup>Created on 2022-06-13 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

simonpcouch:
Awesome—I'm away from work for the week but will give this a more thorough look + merge if things look good next week. :)

Issue #1114, opened 2022-08-22
Tidying `anova()` output from `lme4` models


AmeliaMN:
## The problem

I'm re-running some code that worked before I updated R/all my packages. I'm getting an error from `broom` related to duplicated names in the `anova()` output from `lme4` models. 

## Reproducible example

``` r
  library(tidyverse)
  library(lme4)
#> Loading required package: Matrix
#> 
#> Attaching package: 'Matrix'
#> The following objects are masked from 'package:tidyr':
#> 
#>     expand, pack, unpack
library(broom)
fm1 <- lmer(Reaction ~ Days + (Days | Subject), sleepstudy)
fm2 <- lmer(Reaction ~ Days + (Days || Subject), sleepstudy)
anova(fm1, fm2) %>%
  tidy()
#> refitting model(s) with ML (instead of REML)
#> Warning in tidy.anova(.): The following column names in ANOVA output were not
#> recognized or transformed: npar
#> Error in `mutate()`:
#> ! Can't transform a data frame with duplicate names.
```

<sup>Created on 2022-08-22 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

<details style="margin-bottom:10px;">
<summary>
Session info
</summary>

``` r
sessioninfo::session_info()
#> ─ Session info ───────────────────────────────────────────────────────────────
#>  setting  value
#>  version  R version 4.2.1 (2022-06-23)
#>  os       macOS Monterey 12.2.1
#>  system   aarch64, darwin20
#>  ui       X11
#>  language (EN)
#>  collate  en_US.UTF-8
#>  ctype    en_US.UTF-8
#>  tz       America/Chicago
#>  date     2022-08-22
#>  pandoc   2.18 @ /Applications/RStudio.app/Contents/MacOS/quarto/bin/tools/ (via rmarkdown)
#> 
#> ─ Packages ───────────────────────────────────────────────────────────────────
#>  package       * version    date (UTC) lib source
#>  assertthat      0.2.1      2019-03-21 [1] CRAN (R 4.2.0)
#>  backports       1.4.1      2021-12-13 [1] CRAN (R 4.2.0)
#>  boot            1.3-28     2021-05-03 [1] CRAN (R 4.2.1)
#>  broom         * 1.0.0.9000 2022-08-22 [1] Github (tidymodels/broom@1f017a5)
#>  cellranger      1.1.0      2016-07-27 [1] CRAN (R 4.2.0)
[...]

Issue #1113, opened 2022-08-12
patches needed before Matrix 1.4-2 release


simonpcouch:
From email:

```
Dear R Maintainer maintainer,

You are receiving this message because at least one CRAN or
Bioconductor package that you maintain requires revisions due to
deprecations in the forthcoming Matrix version 1.4-2, which you can
install with

> install.packages("Matrix", repos = "http://r-forge.r-project.org/")

Matrix 1.4-2 will formally deprecate 187 coercion methods.  More
precisely, coercions of the form

> as(object, Class)

where

* 'object' inherits from the virtual class Matrix, is a traditional
  matrix, or is a logical or numeric vector

* 'Class' specifies a non-virtual subclass of Matrix, such as
  dgCMatrix, but really any subclass matching the pattern

      ^[dln]([gts][CRT]|di|ge|tr|sy|tp|sp)Matrix$

will continue to work as before but signal a deprecation message or
warning (_message_ in the widely used dg.Matrix and d.CMatrix cases).
```

simonpcouch:
Closed in [1f017a5](https://github.com/tidymodels/broom/commit/1f017a52aed20acabc58c8ce6762609f7f2b5197). :)

PR #1115, opened 2022-08-23
Anova mermod fix

bbolker:
This addresses https://github.com/tidymodels/broom/issues/1114

### DIFF
--- a/R/stats-anova-tidiers.R
+++ b/R/stats-anova-tidiers.R
@@ -67,6 +67,7 @@ tidy.anova <- function(x, ...) {
     "Sum of Sq" = "sumsq",
     "F" = "statistic",
     "Chisq" = "statistic",
+    "npar" = "npar",
     "P(>|Chi|)" = "p.value",
     "Pr(>|Chi|)" = "p.value",
     "Pr(>Chi)" = "p.value",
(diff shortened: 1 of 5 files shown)


bbolker:
Ugh, needs more work. I made some unwarranted assumptions.

simonpcouch:
Thanks for getting this started!

> Ugh, needs more work. I made some unwarranted assumptions.

I felt this.🫠 Working with this output can be really tricky—I wish we had some more attributes or classes to work with.

Let me know if you'd appreciate me spending some time here! Glad to review whenever feels good to you.

R/stats-anova-tidiers.R
```diff
@@ -127,8 +133,8 @@ tidy.anova <- function(x, ...) {
       ret < cbind(cbind(term, ret), response)
       row.names(ret) <- NULL
     }
-  } else if (length(grep("Model", x_attr$heading)) != 0) {
-    mods <- sub(".*: ", "", strsplit(x_attr$heading[grep("Model", x_attr$heading)], "\n")[[1]])
+  } else if (length(mod_lines) != 0) {
```
simonpcouch, line 136:
```suggestion
  } else if (is.null(ret$term) & length(mod_lines) != 0) {
```

simonpcouch: This would be sufficient to ensure that merMod anova output isn't caught by any of these conditionals, I think.

simonpcouch (commented):
That sounds like a good approach! re: the copying code downside, much of the `tidy.anova` source is now the `Special catch for car::linearHypothesis` conditional (which I presume you wouldn't need?) and I could export the `renamers` object as `anova_tidier_renamers` so that that vector would automatically be synced up. Another downside, though, is that this would require users to load broom.mixed in a place where they used to be able to get by with only the broom package loaded, and the reason for that isn't visible in the sense that output from the `anova()` generic could require either package depending on the method dispatched to.

Another option, which is less expressive but is more backward-compatible and likely more maintainable, suggested in review:

R/stats-anova-tidiers.R
simonpcouch: Unfortunately, I'm a fixed timeline to submit to CRAN due to [upcoming breaking changes from a dependency](https://github.com/tidymodels/broom/issues/1113). I will go ahead and move forward with this suggestion unless you'd like to move forward otherwise!

bbolker: If you've got a fix, that's great - go ahead!

PR #1120, opened 2022-10-02
rm duplicate instance of contributor brunson

corybrunson:
This is a minor change that removes a duplication (of myself) in the list of contributors.


simonpcouch (approved):
Looks great, thanks!

PR #1122, opened 2022-10-13
Migrate `ggplot2` from strong to weak dependency

ddsjoberg:
I may be missing something, but it seems that ggplot2 is only used in documentation, and does not need to be listed in `Imports`. In this PR, I've moved ggplot2 to `Suggests`.

(I was doing some dependency reduction on my own packages, and was wondering how ggplot2 made it into my dependency list.)

### DIFF
--- a/.github/workflows/R-CMD-check-hard.yaml
+++ b/.github/workflows/R-CMD-check-hard.yaml
@@ -50,6 +50,7 @@ jobs:
             any::testthat
             any::knitr
             any::rmarkdown
+            any::ggplot2
           needs: check
 
       - uses: r-lib/actions/check-r-package@v2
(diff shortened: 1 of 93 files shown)


simonpcouch:
Thanks for the PR!

I do agree that migrating this dependency to Suggests is a good move!

This will take a few more changes than the current PR contains; the R-CMD-CHECK-hard check ensures that we use Suggested packages conditionally properly, and thus we ought not to manually specify Suggested packages to be installed as Imports would be there.

I think this is worth the effort, though. :) I'll push some changes here in a moment.

simonpcouch:
Awesome, thanks again for making this happen @ddsjoberg—we ought to have made this change a while ago!

Issue #946, opened 2020-10-08
.resid column not present with variables are log transformed


Amogh-Joshi:
When you are ready to file the bug 🐛 report, please delete everything above this line:
< -- ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ -->

## The problem

I'm having trouble with getting the residual values when we transform the variable. Usually the tibble has 9 columns but for the log transformation, it just provides 8.  This issue is not present when the variable is pre-transformed. 

## Reproducible example

```r
## copy your code to the clipboard and run:
reprex::reprex(si = TRUE)
```
``` r

library(reprex)

library(alr4)
#> Loading required package: car
#> Loading required package: carData
#> Loading required package: effects
#> Registered S3 methods overwritten by 'lme4':
#>   method                          from
#>   cooks.distance.influence.merMod car 
#>   influence.merMod                car 
#>   dfbeta.influence.merMod         car 
#>   dfbetas.influence.merMod        car
#> lattice theme set by effectsTheme()
#> See ?effectsTheme for details.
library(broom)


modelUN <- lm(I(log(fertility)) ~ I(log(ppgdp)), data = UN11)

augment(modelUN)
#> # A tibble: 199 x 8
#>    .rownames `I(log(fertilit~ `I(log(ppgdp))` .fitted .std.resid    .hat .sigma
#>    <chr>             <I<dbl>>        <I<dbl>>   <dbl>   <I<dbl>>   <dbl>  <dbl>
#>  1 Afghanis~            1.79             6.21   1.38    1.34     0.0156   0.306
#>  2 Albania              0.422            8.21   0.965  -1.77     0.00516  0.305
#>  3 Algeria              0.762            8.41   0.924  -0.530    0.00503  0.308
#>  4 Angola               1.64             8.37   0.931   2.30     0.00504  0.304
#>  5 Anguilla             0.693            9.53   0.692   0.00500  0.00739  0.308
#>  6 Argentina            0.776            9.12   0.776  -0.000217 0.00593  0.308
#>  7 Armenia              0.551            8.02   1.00   -1.48     0.00544  0.306
#>  8 Aruba                0.513           10.0    0.586  -0.239    0.0102   0.308
#>  9 Australia            0.667           11.0    0.397   0.889    0.0179   0.307
#> 10 Austria              0.297           10.7    0.445  -0.486    0.0156   0.308
[...]

Issue #937, opened 2020-09-18
.resid column doesn't exist with log output in augment.lm


simonpcouch:
Thanks to @rmtrane for noting the issue and supplying a reprex!

``` r
library(broom)

packageVersion("broom")
#> [1] '0.7.0'

augment(lm(data = mtcars,
           mpg ~ hp))
#> # A tibble: 32 x 9
#>    .rownames           mpg    hp .fitted .resid .std.resid   .hat .sigma .cooksd
#>    <chr>             <dbl> <dbl>   <dbl>  <dbl>      <dbl>  <dbl>  <dbl>   <dbl>
#>  1 Mazda RX4          21     110    22.6 -1.59      -0.421 0.0405   3.92 3.74e-3
#>  2 Mazda RX4 Wag      21     110    22.6 -1.59      -0.421 0.0405   3.92 3.74e-3
#>  3 Datsun 710         22.8    93    23.8 -0.954     -0.253 0.0510   3.92 1.73e-3
#>  4 Hornet 4 Drive     21.4   110    22.6 -1.19      -0.315 0.0405   3.92 2.10e-3
#>  5 Hornet Sportabout  18.7   175    18.2  0.541      0.143 0.0368   3.93 3.89e-4
#>  6 Valiant            18.1   105    22.9 -4.83      -1.28  0.0432   3.82 3.69e-2
#>  7 Duster 360         14.3   245    13.4  0.917      0.250 0.0976   3.92 3.38e-3
#>  8 Merc 240D          24.4    62    25.9 -1.47      -0.396 0.0805   3.92 6.88e-3
#>  9 Merc 230           22.8    95    23.6 -0.817     -0.217 0.0496   3.93 1.23e-3
#> 10 Merc 280           19.2   123    21.7 -2.51      -0.661 0.0351   3.90 7.94e-3
#> # … with 22 more rows

augment(lm(data = mtcars,
           log(mpg) ~ hp))
#> # A tibble: 32 x 8
#>    .rownames         `log(mpg)`    hp .fitted .std.resid   .hat .sigma   .cooksd
#>    <chr>                  <dbl> <dbl>   <dbl>      <dbl>  <dbl>  <dbl>     <dbl>
#>  1 Mazda RX4               3.04   110    3.08    -0.213  0.0405  0.189 0.000958 
#>  2 Mazda RX4 Wag           3.04   110    3.08    -0.213  0.0405  0.189 0.000958 
[...]

Issue #124, opened 2016-05-06
no .resid column with augment.lm and newdata


cderv:
help page says that with newdata argument supplied, 3 columns are added to the dataset. 

> When newdata is supplied, augment.lm returns one row for each observation, with three columns added to the new data:
> - `.fitted`  Fitted values of model
> - `.se.fit` Standard errors of fitted values
> - `.resid` Residuals of fitted values on the new data

However only 2 are there - no `.resid` column is calculated

``` r
library(dplyr)
library(broom)

train <- mtcars %>%
  sample_frac(0.7)
test <- setdiff(mtcars, train)


model <- lm(mpg ~ cyl + disp, data= train)

augment(model, newdata = test) %>% 
  names()
#>  [1] "mpg"     "cyl"     "disp"    "hp"      "drat"    "wt"      "qsec"   
#>  [8] "vs"      "am"      "gear"    "carb"    ".fitted" ".se.fit"
```

So either, we could add a `.resid` column but it implies some calculation or at least the help page should be modified.

double-y:
I also encountered this.

cderv:
I could make a PR on this if it could help...

simonthelwall:
same issue here

topepo:
Same thing here. 

One issue is that `newdata` may not have the outcome variable in it. For `lm` (and most other single numeric outcome models), this can be used to determine if needed variable is present:
```r
lm_response_var <- function(x) {
  y_index <- attr(x$terms, "response")
  if(y_index == 0) return(NA)
  var_list <- attr(x$terms, "variables")
  as.character(var_list[y_index + 1])
}
```
So inside `augment.lm` somewhere, there could be something like
```r
if(!is.na(lm_response_var(x)) || lm_response_var(x) %in% colnames(newdata))
   res$.resid <- blah blah blah
```

alexpghayes:
The goal is to have this behavior for all relevant augment methods, not just `augment.lm`, in 0.7.0. The first step will be adding to this augment test suite.

IndrajeetPatil:
This is fixed in the development version.

``` r
library(dplyr)
#> 
#> Attaching package: 'dplyr'
#> The following objects are masked from 'package:stats':
#> 
#>     filter, lag
#> The following objects are masked from 'package:base':
#> 
#>     intersect, setdiff, setequal, union
library(broom)

train <- mtcars %>%
  sample_frac(0.7)
test <- setdiff(mtcars, train)


[...]

PR #1123, opened 2022-10-13
handle `.resid` in `augment` output when response includes function calls

### LINKED ISSUE #1121: Missing .resid when terms in the formula are function calls
This is a request to reopen #124 , with a potential diagnosis and fix.
This bug describes how the .resid column is missing when terms in the formula are function calls of variables, eg. `log(var)`.

I believe the problem is in the response function in utilities.R; it attempts to create the response variable by calling `model.response` on a new version of the data set, created using `model.frame(terms(object), data = newdata)`. The problem with this is that if `newdata` was created using `model.frame` in the first place (which it is by default in `augment_newdata`), it will already have the results from the function calls instead of the data itself, so recreating the function calls fails.

The solution is to check if newdata already has a response (ie, was created from `model.frame`), and if so, to use that directly, instead of recreating it. If not, then the current code should be run. The below code adds this, wrapping it in a `possibly`, as if it doesn't have a response associated with it, an error is likely.

See ?model.frame/?get_all_vars (the last paragraph of Details) for a brief explanation of what `model.frame` does, in contrast with with `get_all_vars` does. I don't think we can't use `get_all_vars` instead because it requires the raw data set as a parameter.

```
response <- function(object, newdata = NULL) {
  ## if the newdata was created as a model.frame, it will have a response already
  ## wrap this in possibly, because if not, an error is likely
  safe_model.response <- purrr::possibly(model.response, NULL)
  resp <- safe_model.response(newdata)
  ## if not, need to recreate it from the newdata
  if(is.null(resp)) {
    resp <- model.response(model.frame(terms(object), data = newdata, na.action = na.pass))
  }
  resp
}
```

simonpcouch:
Closes #1121, related to #946, #937, #124.🐬


simonpcouch:
Would you be able to take a look at this, @alexpghayes? You mentioned some edge cases to be wary of in #1121, so wanted to give you a chance to see this before merging.

alexpghayes:
Yes, I'll take a look by the end of the week.

alexpghayes:
I just took a look and this seems fine to me! I don't really remember the implementation details as well as I would like, but I find the tests convincing :).

alexpghayes (approved):

simonpcouch:
Much appreciated, @alexpghayes!

PR #1112, opened 2022-08-06
Improve performance of tidy.lm and tidy.glm for full-rank fits

capnrefsmmat:
tidy.lm and tidy.glm use summary(x)$coefficients, then join with coef(x)
to handle the case where the fit is rank deficient and some terms with
NA coefficients are missing from summary(x)$coefficients. However, in a
simple experiment, the left_join() consumes a disproportionate amount of
time. Only run the left_join() when the fit is rank-deficient and it is
necessary, and add tests to ensure the rows are added correctly for
rank-deficient fits.

An example:

```
library(broom)
library(profvis)

profvis(replicate(1000, tidy(lm(dist ~ speed, data = cars))))
```

Before this change, the expression ran in 3470 ms, of which 2570
ms (74%) was in tidy.lm() and 1590 ms (46%) was in
left_join.data.frame(). After this change, the expression ran in 1740
ms, of which 1010 ms was in tidy.lm(), cutting 60% from the total time
in tidy.lm(). This is useful for simulation studies that use tidy() to
get estimates for each simulated fit.

Thanks for making a pull request to broom! A few things to keep in mind that will probably help this PR be merged more quickly:  

* [ ] Have you documented the change in `NEWS.md`?  
  * As this doesn't change any user-facing API, I haven't, but can add it if performance changes are significant
* [x] If this is your first time PRing to broom, have you added yourself as a contributor in the `DESCRIPTION`?
* [x] Have you added any new vocabulary to `inst/WORDLIST`? (New vocabulary will be noted in the R-CMD-check from GitHub Actions.)
* [ ] Does R-CMD-check pass on GitHub Actions? (Sometimes, checks may not be passing on the main branch already—if that's the case, just try to make sure your changes don't add any additional errors/warnings.) 
* [x] Have you updated `DESCRIPTION` if your feature/bug fix requires a specific version of a package?
* [x] Have you added unit tests for any new functionality?

PR #1126, opened 2022-11-06
Take advantage of matrix column support

capnrefsmmat:
I noticed the documentation for `augment.lm()` claimed that matrix columns are not supported and hence `poly()` won't work, but I noticed this while augmenting a model with a `poly()` term that worked just fine. It turns out matrix column support was added in tibble 1.0.0, so matrix columns will work fine. I've updated the augment template documentation accordingly.

Examining the [NEWS for broom 0.5.0](https://github.com/tidymodels/broom/blob/a7d6df53aaf7b95d8949647c75a3a68a506209ca/NEWS.md#breaking-changes), I saw this also meant `augment.survreg()` and `augment.coxph()` required an explicit data argument, because their `model.frame()` would include matrix columns. I verified that using `model.frame()` works just fine now. This should not be a breaking change, because anyone who left the data argument blank before would have gotten an error, whereas now they will get a meaningful result.

I have not committed the updated Rd files, because I don't have all the zillion packages supported by broom installed. But I have verified that the survival tests still pass. You're welcome to rebuild the Rd on this branch.

I'll commit a NEWS update for the `augment.survreg()` and `augment.coxph()` shortly. I'm already in the DESCRIPTION from #1112.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,6 @@
 # broom 1.0.1.9000
 
+* `augment.coxph()` and `augment.survreg()` no longer require `data` or `newdata` arguments to be passed, and use `model.frame()` by default (#1126 by `@capnrefsmmat`).
 * Migrated 'ggplot2' from strong to weak dependency, i.e. moved from `Imports` to `Suggests`.
 * Fixed a bug where `augment()` results would not include residuals when the response term included a function call (#1121, #946, #937, #124).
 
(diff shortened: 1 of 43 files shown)


capnrefsmmat:
The first CHECK failure appears to be because I didn't rebuild the Rd files, as I mentioned, not because of a test failure.

simonpcouch:
Great, thank you @capnrefsmmat! Just pushed those doc updates.

Issue #1132, opened 2022-12-22
migrate (and enhance) zeroinfl and hurdle methods from poissonreg package?


bbolker:
The [poissonreg package](https://github.com/tidymodels/poissonreg) (which is already under the `tidymodels` umbrella) has tidiers for `hurdle` and `zeroinfl` models from the `pscl` package.  These are relatively hard to discover: could they (with permission from the `poissonreg` maintainer) be migrated into `broom`?

In the process it would be good to

- implement a `conf.int` argument 
- implement an `exponentiate` argument (makes sense for both the conditional and zero components, i.e. differences in log-counts/log-fold changes → ratios or fold changes vs. log-odds to odds ratios)

right now these are both silently ignored.

Might submit a PR if encouraged.

hfrick:
Hi Ben, I maintain poissonreg and think broom is a good place for those tidiers to live in. I spoke with Simon and I'll make PRs moving them here (as they currently are) - he'll be the point of contact for any changes to them 👍

bbolker:
Thanks!  I'll leave this open for the enhancement requests (working `conf.int` and `exponentiate` arguments, warning/error when ignoring arguments in `...`)

simonpcouch:
Thanks for the issue, @bbolker! After some further internal discussion, we've decided to leave those tidiers where they're at—feel free to PR those improvements to poissonreg, if you're so inclined.🙂

Generally, in tidymodels, we define methods for objects in the package where the object is defined. In light of your issue, we considered that tidiers for objects defined in non-tidymodels packages that don't define their own tidiers should live in broom. We ultimately decided not to enforce this, as:

* Managing dependencies/namespaces during the transition is a bit gnarly (see #1133), 
* Migrating tidiers here from parsnip extensions where they're needed makes those extensions more indirect-dependency-heavy, and
* We're working on a tool that makes it more easy to find tidiers. broom.mixed, as well as all broom and generics reverse dependencies, will be indexed and searchable in that tool, so hopefully this will ease the process of locating needed tidiers if they don't live in broom or the model-supplying package.

PR #1133, opened 2023-01-03
move pscl tidiers here from poissonreg

hfrick:
part of https://github.com/tidymodels/broom/issues/1132

the corresponding PR on poissonreg is https://github.com/tidymodels/poissonreg/pull/58


hfrick:
The test failures look like they are unrelated to this PR 👀

simonpcouch (approved):
Thank you, @hfrick!

Agreed that the check failures aren't related.

While we'll plan on making some changes to these tidiers before CRAN release, this PR looks good to get that initial transfer done!

hfrick:
Your comment about a smooth transition on the other PR made me remember that we could make the registration of the methods here conditional on the installed version of poissonreg, i.e. if there were already registered in poissonreg (through an older version), don't try to register them again here. I'll push some changes for you to review!

DESCRIPTION
```diff
@@ -551,7 +551,8 @@ Imports:
     rlang,
     stringr,
     tibble (>= 3.0.0),
-    tidyr (>= 1.0.0)
+    tidyr (>= 1.0.0),
+    vctrs
```
hfrick, line 555: this dependency can be removed again when we register the methods unconditionally

R/pscl-tidiers.R
```diff
@@ -0,0 +1,39 @@
...
+#' @param type A character string for which model coefficients to return:
+#' "all", "count", or "zero".
+#' @param ... Not currently used.
+#' @return A tibble
+#'
+#' @name tidy_zip
+tidy_zeroinfl <- .tidy
+
+#' @rdname tidy_zip
+tidy_hurdle <- .tidy
```
hfrick, line 39: I've left this so that we only need to change the `_` back to the `.` when we register unconditionally (instead of using `.tidy()` directly in the registration in `.onLoad()`.

simonpcouch:
Closing, with some background in [this comment](https://github.com/tidymodels/broom/issues/1132#issuecomment-1398854168).🐙

PR #1141, opened 2023-02-04
Fixed survexp test to no longer use ratetable function

dgrtwo:
I received an email from Dr. Therneau, who maintains the survival package:

> One of the test files in broom uses a call to survexp that is out of date, and as of the next release of the package will no longer be supported.    Essentially, you have a very old copy of an example in survexp.Rd.    I've attached an updated file, to make it easier to update.   (Per my mercurial logs, I think it changed in 2018).

The change is his (but works for me locally)


simonpcouch:
Thanks for passing this along! Looks good to me.

PR #1144, opened 2023-02-28
Allow tidy.aov to return intercept data

victor-vscn:
Changes tidy.aov on stats-anova-tidiers.R to allow an argument called intercept which activates the same argument on summary.aov. This results in a new top row with intercept information, which is complete (MS, SS, t, p). This is similar to the output in many statistical software.

My proposal could also be accomplished by resolving the dots on the summary function, as they're not currently being used, but that's a larger modification (in terms of consequences) I'm not comfortable doing at this time.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,7 @@
 # broom (development version)
 
+* Added an `intercept` argument to `tidy.aov()`, a logical indicating whether to include information on the intercept as the first row of results (#1144 by `@souza-victor`).
+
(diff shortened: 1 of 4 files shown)


simonpcouch:
Thanks for the PR, @souza-victor! Just pushed a few changes, will revdepcheck and see if this affects any packages downstream before merging.

simonpcouch:
You got it, @victor-vscn. :)

PR #1156, opened 2023-05-17
Make id column names agnostic in augment.mlogit

gregmacfarlane:
This fixes an issue where mlogit can give different names to the id columns in the estimation dataset.


simonpcouch:
Could you provide an example demonstrating the issue that this PR addresses? I'm hesitant to defer to column positions without understanding the problem we're solving a bit better.

gregmacfarlane:
This issue is described in #1157 . An index is safe in this case, because everything in mlogit works off of these indices in the appropriate order.

simonpcouch:
Thanks so much for the additional context! You got it—just pushed some testing and a NEWS entry.

PR #1172, opened 2023-08-30
Added adj.r.squared and npar as outputs of glance.gam for mgcv::gam

Tripartio:
When using glance.gam for mgcv::gam, I was unpleasantly surprised to not find any results for adjusted R squared. This is one of the most important model evaluation statistics for statistical inference. Not only my colleagues and I, but also probably most other users of glance.gam would need this.

On examining the code, it seems that the only outputs provided by the current glance method are those directly available from the gam model object. But summary.gam provides some additional valuable outputs. So, I have modified the code to calculate the summary(x) on the model object (x) and then add in the other useful outputs (at least, those, that output a scalar numeric output.

That said, when examining modeltests::column_glossary, I could not find most of the additional outputs. So, I have added the only two outputs available in the modeltests::column_glossary (adj.r.squared and npar); for the other outputs, I have listed them in the code but commented them out--that way, if they are supported in the future in modeltests::column_glossary, the code can more easily be updated.

I have tested the updates with the unit tests in tests/testthat/test-mgcv.R; my modifications pass the tests. I hope you can accept this pull request.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,6 +2,8 @@
 
 * Moved forward with deprecation of tidiers for objects from the sp package. See resources linked in [tidymodels/broom#1142](https://github.com/tidymodels/broom/issues/1142) for more information on migration from retiring spatial packages.
 
+* Added support for columns `adj.r.squared` and `npar` in `glance()` method for objects outputted from `mgcv::gam()` (#1172).
+
 # broom 1.0.5
 
 * `tidy.coxph()` will now pass its ellipses `...` to `summary()` internally (#1151 by `@ste-tuf`).
--- a/R/mgcv-tidiers.R
+++ b/R/mgcv-tidiers.R
@@ -74,11 +74,11 @@ tidy.gam <- function(x, parametric = FALSE, conf.int = FALSE,
     class(sx) <- c("anova", "data.frame")
     ret <- tidy(sx)
   }
-
+  
   if (exponentiate && parametric) {
     ret <- exponentiate(ret)
   }
-
+  
   ret
 }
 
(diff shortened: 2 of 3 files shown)


simonpcouch:
Thanks for the PR! This looks solid.

Wanted to make sure we wouldn't be increasing the time-to-tidy too drastically by introducing the additional `summary()` call, but looks like that's not an issue:

``` r
library(mgcv)
#> Loading required package: nlme
#> This is mgcv 1.9-0. For overview type 'help("mgcv-package")'.
library(broom)

set.seed(2) ## simulate some data... 
dat <- gamSim(1, n = 400, dist = "normal", scale = 2)
#> Gu & Wahba 4 term additive model

b <- gam(y ~ s(x0) + s(x1) + s(x2) + s(x3), data = dat)

bench::mark(
  gam = gam(y ~ s(x0) + s(x1) + s(x2) + s(x3), data = dat),
  tidy = tidy(b),
  summary = summary(b),
  check = FALSE,
  relative = TRUE
)
#> # A tibble: 3 × 6
#>   expression   min median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr> <dbl>  <dbl>     <dbl>     <dbl>    <dbl>
#> 1 gam        36.1   36.1       1         34.9     1   
#> 2 tidy        4.62   4.57      7.78      32.8     2.78
#> 3 summary     1      1        35.8        1       2.67
```

<sup>Created on 2023-09-06 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

Pushing some small changes before merging.

PR #1180, opened 2023-12-04
Use testthat 3

olivroy:
I also silenced a couple of tests / output along the way.


R/stats-kmeans-tidiers.R
```diff
@@ -62,7 +62,7 @@ augment.kmeans <- function(x, data, ...) {
 
   # kmeans allows for input matrices without column names,
   # so add them in the same way that fix_data_frame() would have
-  if (inherits(data, "matrix") & is.null(colnames(data))) {
+  if (inherits(data, "matrix") && is.null(colnames(data))) {
```
simonpcouch, line 65: Could you pull these changes from `&` to `&&` out into a separate PR, clarifying the bugs you see arising from the current implementation?

simonpcouch (changes requested):
Thank you for the PR! This is a lot of work; I appreciate you putting the time in here.

Could you separate the test changes, vectorized `rlang::is_installed()`, `name_repair` silencing, and transitions from `&` to `&&` in internals into separate PRs? That will help me review more informatively.

R/stats-kmeans-tidiers.R
olivroy: Opened #1182, #1181. I will update this PR when they are merged.

olivroy:
Thanks for triggering CI. As you see, there is still a lot of noise in the [logs](https://github.com/tidymodels/broom/actions/runs/7107452118/job/19348886396?pr=1180#step:6:177). I will try silencing most of it. It is mostly due to the fact that packages loaded in broom's tests are quite noisy on load.

simonpcouch:
Noise in the logs is okay! What's more important is my ability to understand the change set comprehensively. As-is, there are still many changes that seem unrelated to the switch to testthat 3e—I'd welcome those other changes in separate PRs where their scope is more clearly defined, but will ask you to further pare back the changes here to only those needed to related to switching to 3e.

PR #1184, opened 2023-12-05
Fixup package docs

olivroy:
Remove duplicated entry


DESCRIPTION
simonpcouch, line 1: Could you revert the changes in this file?

R/broom-package.R
```diff
@@ -1,13 +1,17 @@
...
+# Custom package description for the package help file.
+#' @description
+#' Convert statistical analysis objects from R into tidy tibbles,
+#' so that they can more easily be combined, reshaped and otherwise processed
+#' with tools like dplyr, tidyr and ggplot2. The package provides three S3
+#' generics: tidy, which summarizes a model's statistical findings such as
+#' coefficients of a regression; augment, which adds columns to the original
+#' data such as predictions, residuals and cluster assignments; and glance,
+#' which provides a one-row summary of model-level statistics.
+#'
```
simonpcouch, line 10:
```suggestion
#' 
#' @aliases broom broom-package
#' 
```

So that folks can still `?broom` and that dead link will still work.

olivroy: As you can see in the .Rd file, it still keeps the broom alias. So ?broom will work.

For keeping the broom link, I can add `@name broom`

olivroy:
Comments addressed. The fixes won't break the link!

simonpcouch (approved):
Looking good. :)

Issue #949, opened 2020-10-11
Confidence level in augment.lm


julian-urbano:
Hi!

I was checking `augment` to compute intervals, and wanted to use different confidence levels. I presumed this would be done via something like `conf.level=0.9`, thinking about how `predict` is used. However, the intervals are always the same, regardless of the confidence level.

I looked at the code for `predict.lm` and found this line here https://github.com/tidymodels/broom/blob/8bd38356c2f675f6889df2627d687d110a4c0fe7/R/stats-lm-tidiers.R#L136

I think it should read as follows
```r
df <- augment_newdata(x, data, newdata, se_fit, interval, ...)
```
to forward other arguments, such as `conf.level` in this case.

Thoughts?

simonpcouch:
This sounds like a good idea to me! 

re: the comment `be *incredibly* careful that the ... are passed correctly` in the `augment_newdata` comments, checking in—does this sound good to you, @alexpghayes?

petebaker:
Hi @julian-urbano, @simonpcouch and @alexpghayes 

I would also like to nudge this topic along but I'm not sure about the interest level. I think that _broom_ is fantastic and it's a great step in the right direction. However, being able to set the `level` for a CI but not a PI is a deal breaker for me and I believe it is a trap for the unwary.

Recently, I **highly recommended** _broom_ to colleagues but once I found out the state of play w.r.t. prediction intervals I've **highly un-recommended** it. Here's why.

My colleagues wanted to study various prediction intervals  80%, 90% etc but they pointed out that the prediction intervals are always 95% which makes `augment.lm` essentially useless for them, and unfortunately they couldn't work out how to make it work so they abandoned `augment.lm`  and rolled their own using `predict.lm` and `level`.

The documentation around `...` is somewhat messy but the important problem is that `...` is not implemented. The documentation gives a heads up that things can go wrong when options are chosen via `...` which is perfectly sensible. (However, given that  `...` isn't implemented for prediction, then it is simply wrong). Am I missing something here?

I think _broom_ is great but if standard options/arguments like `level` for `predict.lm` simply aren't implemented when it appears that they are available, then this seems dangerous to me.

If maintainers are interested, then I will have a go at a pull request.

Comments most welcome!

Cheers
Peter

PS: Having produced both CIs and PIs for about 40 years now, then I would say that prediction intervals for new observations are often more useful than confidence intervals for the mean response (unless fixed at some arbitrary level of course).

ptoche:
[...]

PR #1191, opened 2024-03-07
Add support for `conf.level` in `augment.lm`

zietzm:
Addresses https://github.com/tidymodels/broom/issues/949.

https://github.com/alexpghayes/modeltests/pull/38 added support for the `conf.level` argument in `augment` methods. 

This PR adds support for user-specified `conf.level` in `augment.lm`, which previously returned only values for `conf.level=0.95`, without any user alert.


simonpcouch (commented):
I'm feeling hesitant to make changes to `augment_newdata()`, which underlies augmenting for many other methods in the package. I do wonder if we may be better off passing `augment.lm`'s dots to `augment_newdata()`'s, as `augment.lm()`'s dots are currently unused and not checked for emptiness--this would allow for passing `conf.level` without requiring any changes to `augment_newdata`.

Open to your opinions, @alexpghayes.

alexpghayes:
Yeah, so my general philosophy is that all the utilities like `augment_newdata()` are enormous maintenance liabilities and the best move is to get rid of them. This seems like an excellent time to factor out the dependence of `augment.lm()` on `augment_newdata()`. I would just copy-paste the `augment_newdata()` implementation into `augment.lm()` and then apply these changes within the `augment.lm()` method only.

I know this approach sounds like it creates a ton of copy-pasted code, but I swear it is the right move for broom. Every model object behaves a little bit differently, and the best way to handle this is for each model object to have genuinely custom/standalone tidier implementations that don't rely on things like `augment_newdata()`.

zietzm:
Updated with @alexpghayes copy-paste suggestion. I tried to make only minimal changes to the pasted code. Happy to clean it up more if desired.

simonpcouch:
Pasting the `augment_newdata()` definition into `augment.lm()` doesn't feel much like a win for maintainability to me. There are a few places in the `augment_newdata()` definition that are accommodating edge cases for other model objects and the process of trimming those out for only what's needed for `lm` sounds labor-intensive and error-prone.

Thanks for the quick updates, @zietzm. If you're game for just passing `augment.lm()`'s dots to `augment_newdata()` and documenting that those dots are now passed to `predict.lm()` I'll be game to review! Will go ahead and enable Actions.

zietzm:
@simonpcouch, either approach works for me! This is certainly the simpler approach. Updated the PR and happy to make any more changes as you see fit.

simonpcouch:
Feel free to ignore the actions dependency errors; unrelated and looking into them right now.

zietzm:
@simonpcouch, thanks for your help! Are any further changes needed here now that all tests are passing?

simonpcouch:
Thanks again for the attention here, @zietzm! Every draft of yours looked solid. For consistency with the analogous argument in `tidy()` methods, I went ahead and made use of your work in alexpghayes/modeltests#38 and transitioned that argument name from `predict()`'s `level` to broom's usual `conf.level`. We can depend on a development version of modeltests until a new version makes it to CRAN. :)

Issue #1198, opened 2024-04-25
Allow choosing the type of residuals in `tidy.betareg()`


DanChaltiel:
Hi,

When fitting a beta regression using `betareg::betareg()`, you often have to choose a different type of residual as the default one may produce `NaN`.

For instance, you could consider the following reprex, taken from [StackOverflow](https://stackoverflow.com/a/49140732/3888000).
The `type` argument in `summary()` is forwarded to `residuals()`.

``` r
# here are the data
devtools::source_gist("169bfa3a6c709fd2fd31c5bfa46648ee")
library(betareg)
dieet$Percentage <- gsub(",",".",dieet$Percentage)
dieet$Percentage <- as.numeric(dieet$Percentage)
model.beta = betareg(Percentage ~ Kuikenweek,data = dieet)

summary(model.beta)
#> Warning in sqrt(v * (1 - hatvalues(object))): NaNs produced
#> Call:
#> betareg(formula = Percentage ~ Kuikenweek, data = dieet)
#> Standardized weighted residuals 2:
#> Error in quantile.default(x$residuals): missing values and NaN's not allowed if 'na.rm' is FALSE

summary(model.beta, type = "deviance")
#> 
#> Call:
#> betareg(formula = Percentage ~ Kuikenweek, data = dieet)
#> 
#> Deviance residuals:
#>     Min      1Q  Median      3Q     Max 
#> -2.6188 -0.4916  0.0198  0.4973  2.9149 
#> 
#> Coefficients (mean model with logit link):
#>                   Estimate Std. Error z value Pr(>|z|)    
#> (Intercept)        -0.5872     0.3003  -1.955 0.050542 .  
#> KuikenweekWeek -2   0.1738     0.4208   0.413 0.679594    
#> [..]

tibble::tibble(
  pearson=residuals(model.beta, type = "pearson"),
  deviance=residuals(model.beta, type = "deviance"),
  response=residuals(model.beta, type = "response"),
  weighted=residuals(model.beta, type = "weighted"),
  sweighted=residuals(model.beta, type = "sweighted"),
  sweighted2=residuals(model.beta, type = "sweighted2")
)
#> Warning in sqrt(v * (1 - hatvalues(object))): NaNs produced
#> # A tibble: 36 × 6
#>    pearson deviance response weighted sweighted sweighted2
#>      <dbl>    <dbl>    <dbl>    <dbl>     <dbl>      <dbl>
[...]

PR #1199, opened 2024-05-03
Pass ellipsis to `summary.betareg()`

DanChaltiel:
This PR introduces passing ellipsis to `summary.betareg()`.

This allows choosing the type of residual, as the default one may produce NaN, as explained in detail in #1198.

I'll leave running the documentation to you, as it would require me to install a tremendous amount of packages 😅

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -6,6 +6,8 @@
 
 * Corrected confidence interval values for precision components in `tidy.betareg()` output (#1169).
 
+* `tidy.betareg()` will now pass its ellipses `...` to `summary()` internally (#1198 by `@DanChaltiel`). Among other things, this allows choosing the type of residuals.
+
 * Fixed bug in tidier for `car::linearHypothesis()` output with long formulas (#1171).
 
 * Added support for columns `adj.r.squared` and `npar` in `glance()` method for objects outputted from `mgcv::gam()` (#1172).
--- a/tests/testthat/test-betareg.R
+++ b/tests/testthat/test-betareg.R
@@ -32,6 +32,24 @@ test_that("tidy.betareg", {
   )
 
   check_dims(td1, 12, 8)
+  
+  #test passing ellipsis to summary
+  data_test <- structure(
+    list(p = c(0.4238, 0.8248, 0.5927, 0.3208, 0.8317, 
+               0.7314, 0.4624, 0.5224, 0.3528, 0.7739, 0.1264, 0.4516, 0.4306, 
+               0.5243, 0.5117, 0.149, 0.3342, 0.5069, 0.7101, 0.8019, 0.7569, 
+               0.9096, 0.9013, 0.5403, 0.2264, 0.5775, 0.7366, 0.4086, 0.575, 
+               0.6623, 0.7758, 0.1517, 0.7587, 0.3247, 0.7463, 0.6325), 
+         w = structure(c(1L, 2L, 9L, 4L, 3L, 5L, 1L, 9L, 4L, 3L, 5L, 2L, 4L, 5L, 1L, 2L, 9L, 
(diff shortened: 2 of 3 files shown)


NEWS.md
```diff
@@ -6,6 +6,8 @@
 
 * Corrected confidence interval values for precision components in `tidy.betareg()` output (#1169).
 
+* `tidy.betareg()` will now pass its ellipses `...` to `summary()` internally (#1198 by `@DanChaltiel`). Among other things, this allows choosing the type of residuals.
```
simonpcouch, line 9: I'm a bit confused by the reference to residuals here, as that seems more relevant to predictions with `augment()` than terms with `tidy()`. With this PR (as ought to be the case), passing `residuals` doesn't affect output since `tidy()` doesn't report on residuals:

``` r
library(broom)
library(betareg)
  
data("GasolineYield")
  
fit1 <- betareg(yield ~ batch + temp, data = GasolineYield)

all.equal(
  tidy(fit1, residuals = "deviance"),
  tidy(fit1, residuals = "pearson")
)
#> [1] TRUE
```

<sup>Created on 2024-05-06 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>

tests/testthat/test-betareg.R
```diff
@@ -32,6 +32,24 @@ test_that("tidy.betareg", {
...
+         w = structure(c(1L, 2L, 9L, 4L, 3L, 5L, 1L, 9L, 4L, 3L, 5L, 2L, 4L, 5L, 1L, 2L, 9L, 
+                         4L, 3L, 5L, 8L, 7L, 6L, 1L, 2L, 9L, 3L, 5L, 8L, 7L, 1L, 2L, 9L, 
+                         4L, 3L, 5L), 
+                       levels = c("1", "2", "3", "4", "5", "6", "7", "8", "9"), 
+                       class = "factor")), 
+    row.names = c(NA, -36L), class = c("tbl_df", "tbl", "data.frame")
+  )
+  m <- betareg::betareg(p ~ w, data = data_test)
+  expect_warning(tidy(m))
+  expect_silent(tidy(m, type = "sweighted"))
```
simonpcouch, line 52: Rather than inlining data and making a new fit here, could you just use `fit1` from above, pass an argument through `tidy()`'s `...` that induces a change in output, and `expect_*()` that that output changes?

NEWS.md
DanChaltiel: As you can see in the test file, the type of residuals can be changed using the `type` argument, not `residuals`. I assumed this is the most common argument passed to `summary()`, but I admit it is only based on my own experience.

tests/testthat/test-betareg.R
DanChaltiel: The point of changing the type of residual is that the default one (`sweighted2`) may produce NaN, as explained in detail in https://github.com/tidymodels/broom/issues/1198. 
This inline data is the minimal example I could find to reproduce this behavior. Isn't that right?

NEWS.md
simonpcouch: Ah, my mistake! Same goes for the argument you'd referred to, though:

``` r
library(broom)
library(betareg)

data("GasolineYield")

fit1 <- betareg(yield ~ batch + temp, data = GasolineYield)

all.equal(
  tidy(fit1, type = "deviance"),
  tidy(fit1, type = "pearson")
)
#> [1] TRUE
```

<sup>Created on 2024-05-09 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>

DanChaltiel: Indeed, I must have brainfarted on this whole issue as you are right, the output of tidy is totally independent of the residuals type.
The only interest left is to silence the warning, as in the test below. It makes it clear that you cannot test the output, only the warning.
Not sure if this PR is worth it for this sole use...

simonpcouch: All is well! I totally did as well when I first read the issue. Thanks for the PR—I'll go ahead and close.

Issue #1208, opened 2024-07-01
Glance.lm gives does not give fstatistic/ f stat based values when used on a 0 intercept model. 


jrob95:
## The problem
I have attempted to `glance` a `lm` of the form `y ~ x + 0`. When doing so, `statistic`, `p.value` & `df` are returned as NA.

Examination of the package has led me to believe the issue is in line 213 of `stats-lm-tidiers.R` where glance checks if the model is fitted to intercept only, but it only checks if the summary of the model has one row, which it also does when a 0 intercept model is estimated. 

## Reproducible example

```r
library(broom)

data <- mtcars

model <- lm(data, formula = mpg ~ 0 + cyl)

glance(model)
#> # A tibble: 1 × 12
#>   r.squared adj.r.squared sigma statistic p.value    df logLik   AIC   BIC
#>       <dbl>         <dbl> <dbl>     <dbl>   <dbl> <dbl>  <dbl> <dbl> <dbl>
#> 1     0.734         0.725  11.0        NA      NA    NA  -122.  247.  250.
#> # ℹ 3 more variables: deviance <dbl>, df.residual <int>, nobs <int>

Created on 2024-07-01 with [reprex v2.1.0](https://reprex.tidyverse.org/)
```

simonpcouch:
Closed in #1209. Thank you!

PR #1209, opened 2024-07-01
Update `glance` to correctly function on `lm` models with a 0 intercept

jrob95:
Response to issue: #1208

Have implemented fix, & associated tests and function works as intended.


simonpcouch:
Thanks for the thorough issue description! These changes look clean—just added a NEWS bullet and will pick this up in the morning once checks have run.

